PR #293, opened 2014-12-11
Pagination

mpath:
Enable pagination for networks, subnets and ports.

RM#9187


quark/tests/functional/db/test_pagination.py
```diff
@@ -0,0 +1,132 @@
...
+class QuarkNetworksPaginationFunctionalTest(BaseFunctionalTest):
+    def test_networks_pagination(self):
+        networks_per_page = 1
+        networkA = dict(name="fake_A", tenant_id="fake", network_plugin="BASE")
+        networkB = dict(name="fake_B", tenant_id="fake", network_plugin="BASE")
+        db_api.network_create(self.context, **networkA)
+        db_api.network_create(self.context, **networkB)
+        res = network_api.get_networks(self.context, networks_per_page,
+                                       [('id', 'asc')], None, False,
+                                       None)
```
asadoughi: Just doing a cursory glance at this trying to understand it better. Does this mean neutronclient is sending in the sorting key on every GET request? IIRC, having no sorting key returns unpaged results.

mpath: On line 39, the sorting key is by quark_networks.id ascending. This is one of the functional tests you requested. You do recall correctly, but no, it's not sending the sorting key in every request, just this test.

asadoughi: Right, so, if neutronclient is not sending the sorting key on every GET request, how will this look like when it works in production?

asadoughi: Is there a default sorting key somewhere?

mpath: The user provides the sorting key in the request uri

mpath: So you want a default sort key?

asadoughi: So, the client provides the sorting key in the URI? Is that a neutronclient argument? Or, one we can add to rackspace-python-neutronclient? I assume neutronclient has an idea of pagination built in, maybe? 

Well, a default sort key in quark would make sense if neutronclient doesn't pass that in by default. I'll let you research the tradeoffs and make the decision for whatever makes sense for it to work in production. Thanks.

mpath: https://review.openstack.org/#/c/19926/

It's in neutronclient already.

asadoughi: Sorry, that's +500 lines, can you dig in to a specific line number where the call is made and the sort key is passed in? Have you tested this locally by standing up a dummy neutron-server and making requests with neutronclient? I'm just looking for a description of how this will work in production when a request is made so I can understand this PR better.

mpath: I'd rather not provide a default (yet). Let's see what people actually sort by in production and revisit the idea of defaults when we have some data to base decisions on.

From a user perspective, not having to pass a sort key would be nice, but having a default that sorts by the "wrong" key or in the "wrong" direction would be just as good as not having a default, because they still need to pass the desired values to override the defaults.

TL;DR: Let's leave out the defaults until we know what users are sorting by.

asadoughi: If it's not clear what I am asking or you need help in figuring out the points of integration, feel free to ping me on IRC. I think this thread has gone long enough on github.

asadoughi: To summarize our IRC conversation, you use neutronclient to pass `-P 1 --sort-key=id --sort-dir=asc`.

quark/tests/functional/db/test_pagination.py
```diff
@@ -0,0 +1,132 @@
...
+        ports_per_page = 1
+        with self._stubs(network, [subnet_A_info, subnet_B_info]) as (
+                net, sub_ports):
+            port_api.create_port(self.context, _make_body())
+            port_api.create_port(self.context, _make_body())
+            res_ports = port_api.get_ports(self.context, ports_per_page,
+                                           [('id', 'asc')], None)
+            self.assertEqual(len(res_ports), ports_per_page)
+            res_ports = port_api.get_ports(self.context)
+            self.assertNotEqual(len(res_ports), ports_per_page)
```
asadoughi: Would you mind adding a test case for when sort is not passed in to ensure that ports are sorted by created_at by defautl?

asadoughi:
Finished reviewing latest draft. Just one more test case, please.

mpath:
Jarvis: run tests

jmeridth:
:+1:

PR #300, opened 2014-12-31
Fix KeyError when VM removed

asadoughi:
When security groups are modified, it could be because the VM was
deleted. In that case, when security groups are processed, we need to
catch the exception when the vif expects the device_id to be present.

RM9115


insequent:
My Xen API knowledge is well, zero, but for what it's worth, this LGTM. I definitely want/need more practice in this realm, but the change seems sound.

:snowman:

PR #302, opened 2015-01-07
Atomically update next_auto_assign_ip

Cerberus98:
RM10879

Modifies quark IPAM to atomically update the next_auto_assign_ip field
on subnets without using row locks. The previous implementation was
racy, allowing for the code to set explicit values rather than
incrementing the column as if it were a counter entirely in SQL.

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -476,9 +476,41 @@ def subnet_find_ordered_by_most_full(context, net_id, **filters):
         query = query.filter(models.Subnet.ip_version == filters["ip_version"])
     if "segment_id" in filters and filters["segment_id"]:
         query = query.filter(models.Subnet.segment_id == filters["segment_id"])
+    query = query.filter(models.Subnet.next_auto_assign_ip != -1)
+
     if "subnet_id" in filters and filters["subnet_id"]:
         query = query.filter(models.Subnet.id.in_(filters["subnet_id"]))
+    return query
+
+
+def subnet_update_next_auto_assign_ip(context, subnet):
+    query = context.session.query(models.Subnet)
+    query = query.filter(models.Subnet.id == subnet["id"])
     query = query.filter(models.Subnet.next_auto_assign_ip != -1)
+
+    # For details on synchronize_session, see:
+    # http://docs.sqlalchemy.org/en/rel_0_8/orm/query.html
+    query = query.update(
+        {"next_auto_assign_ip":
+         models.Subnet.next_auto_assign_ip + 1},
+        synchronize_session=False)
+
+    # Returns a count of the rows matched in the update
+    return query
+
+
+def subnet_update_set_full(context, subnet):
+    query = context.session.query(models.Subnet)
+    query = query.filter_by(id=subnet["id"])
+    query = query.filter(models.Subnet.next_auto_assign_ip != -1)
+
+    # For details on synchronize_session, see:
+    # http://docs.sqlalchemy.org/en/rel_0_8/orm/query.html
+    query = query.update(
+        {"next_auto_assign_ip": -1},
+        synchronize_session=False)
+
+    # Returns a count of the rows matched in the update
     return query
 
 
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -669,60 +669,68 @@ class QuarkIpam(object):
                                 segment_id=segment_id, subnet_ids=subnet_ids,
                                 ip_version=filters.get("ip_version"))))
 
-        subnets = db_api.subnet_find_ordered_by_most_full(
-            context, net_id, segment_id=segment_id, scope=db_api.ALL,
-            subnet_id=subnet_ids, **filters)
-
-        if not subnets:
-            LOG.info("No subnets found given the search criteria!")
-
-        for subnet, ips_in_subnet in subnets:
-            ipnet = netaddr.IPNetwork(subnet["cidr"])
-            LOG.info("Trying subnet ID: {0} - CIDR: {1}".format(
-                subnet["id"], subnet["_cidr"]))
-            if ip_address:
-                na_ip = netaddr.IPAddress(ip_address)
-                if ipnet.version == 4 and na_ip.version != 4:
-                    na_ip = na_ip.ipv4()
-                if na_ip not in ipnet:
-                    if subnet_ids is not None:
-                        LOG.info("Requested IP {0} not in subnet {1}, "
-                                 "retrying".format(str(na_ip), str(ipnet)))
-                        raise q_exc.IPAddressNotInSubnet(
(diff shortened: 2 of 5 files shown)


Cerberus98:
Up for feedback. I'm assuming there will be questions

quark/db/api.py
```diff
@@ -476,9 +476,41 @@ def subnet_find_ordered_by_most_full(context, net_id, **filters):
...
+
     if "subnet_id" in filters and filters["subnet_id"]:
         query = query.filter(models.Subnet.id.in_(filters["subnet_id"]))
+    return query
+
+
+def subnet_update_next_auto_assign_ip(context, subnet):
+    query = context.session.query(models.Subnet)
+    query = query.filter(models.Subnet.id == subnet["id"])
     query = query.filter(models.Subnet.next_auto_assign_ip != -1)
```
asadoughi: Shouldn't this be `query = query.filter(models.Subnet.next_auto_assign_ip == current_next_auto_assign_ip)` where `current_next_auto_assign_ip` is a function argument?

asadoughi: Actually, you have the `subnet` model instance there already, so `current_next_auto_assign_ip = subnet.next_auto_assign_ip`.

Cerberus98: I think that actually ends up being more racy than the previous implementation. The reason is everything is trying to update the value EXACTLY as it expects. Under high load, N threads will all try to increment the counter assuming the same next_auto_assign_ip and N-1 will fail

After dong this, the revelation I had is that we should be treating this like a capsule toy machine: instead of getting in line to get a ticket for an IP, you're pulling a lever and dropping an IP into a bin. It doesn't matter which IP fell out of the machine when you pulled the lever as long as you got one when you reach into the tray

This approach approximates that, along with the select for update on subnets, by forcing the IPs to fall in the bin and then be consumed in lock step. I have an idea for how to extrapolate the approach above so that the lock can be removed altogether, but under test this performed very well.

I ran a 1000 port create test using 75 distinct python processes, all generating NEW IPs in the subnet, and this was the outcome:

<pre>
Passed: 1000
Failed: 0
Average: 18.8431840098
Minimum: 11.2398941517
Maximum: 26.5364151001
Stddev: 5.07319573964
Total time: 260.6582129
</pre>

I also confirmed that no IPs were skipped in the subnet. Repeating the test multiple times presented the same results.

asadoughi: I like the new mental model of a dispenser, but it's not clear to me how that's reflected in the proposed implementation here.

quark/ipam.py
```diff
@@ -669,60 +669,68 @@ def select_subnet(self, context, net_id, ip_address, segment_id,
...
+                            updated = db_api.subnet_update_next_auto_assign_ip(
+                                context, subnet)
+
+                        if updated:
+                            context.session.refresh(subnet)
+                        else:
+                            # This means the subnet was marked full while
+                            # we were checking out policies. Fall out and
+                            # go back to the outer retry loop.
+                            return
```
asadoughi: Specifically, can you elaborate a bit more on what's happening here on L719-725.

quark/db/api.py
Cerberus98: Well, you have to use your imagination a little bit on this pass.

Pretend that there's a security guard on the lever (the select for UPDATE). He's making sure people are only pulling that lever one at a time, rather than some insane, Shrodinger level lever pulling. Because he's making sure everyone is going through single file, only one process at a time  (for the most part) is trying to get any given IP.

(The original comment said the select for update was removed. I got ahead of myself)

If I removed the select for UPDATE in this patch, we still skipped IPs.

What I *want* to do next is try one of two things:

* Insert the IP address atomically at the same time as updating the subnet OR
* Pull the lever and then fall out to the reallocate logic.

I like the first way because it's easy to think about. I like the second way because there's only one way to assign an IP. I'm still debating, and I had a 3 week vacation in the middle so I need to collect my thoughts on it all again.

quark/ipam.py
Cerberus98: Sure

Updated is True if the query matched a row and successfully updated it. It's False if it couldn't do that.

The refresh is essentially the same thing as removing the object from the session then doing a SELECT * from subnets where id = X. The reason we do it is the all-in-one subnet_update_next_auto_assign_ip updates the TABLE but not the OBJECT; we bypassed the ORM in that call.

Finally, if we didn't update anything, we need to go back out, pull the lever again, and try to get an IP. The reason it wouldn't update is a) the subnet was deleted or b) The subnet was marked as full (next_auto_assign_ip == -1) by another competing thread.

As above, you have to use your imagination a little bit on this cut of the patch. You've pulled the lever, you go to reach in the bin, and nothing is there, so you need to go back to the end of the line and try it again. The security guard is making sure you aren't going out of turn.

I'm happy to try to add better comments to anything here based on feedback.

Cerberus98:
For reference, I consider this patch the first cut in a series of fixes for the entire IPAM code base. In testing, it looks like it fixes the TOCTOU, which is the main aim of the bug. It does NOT eliminate the chance for deadlocks. I hope to address that with the next version. I didn't want to crank out a gigantic, obtuse patch and hope to get it by anyone, so I'm going to do this piecemeal

asadoughi:
I'm still processing this once over today. Thanks for your patience.

quark/ipam.py
asadoughi:
```
Updated is True if the query matched a row and successfully updated it. It's False if it couldn't do that.
```

Ah, that's helpful to know, but I don't think that matches the comment on the else statement:
```python
 # This means the subnet was marked full while
 # we were checking out policies. Fall out and
 # go back to the outer retry loop.
`

**Comment by asadoughi** [comment]
LGTM. Looking at the tox coverage from travis-ci, It would be nice to have 2 UT around `select_subnet` where updated = 0 as returned from (1) `db_api.subnet_update_set_full` and (2) `db_api.subnet_update_next_auto_assign_ip`.

**Comment by Cerberus98** [comment]
Thanks for reviewing. I'll go add those tests.

**Comment by Cerberus98** [comment]
Updated with two tests for subnet_update_set_full and next_auto_assign_ip. Working on adding select_subnet tests

**Comment by asadoughi** [comment]
```
quark.ipam                                  415     13    97%   86, 172-174, 315-320, 344-345, 694, 725, 848-851
```
694 and 725 are lines in `select_subnet` still not being covered. (The 2 additional tests currently proposed have `updated > 0`)

Cerberus98:
The two current tests wouldn't touch that missing coverage. They're both functional tests I needed to add called in db_api directly. I'm working on a bunch of new select_subnet tests right now

Cerberus98:
Updated with new tests

asadoughi:
LGTM

PR #304, opened 2015-01-12
Set ip_addreses extension to use QuarkIpamANY

TMaddox:
The ip_addresses extension should use QuarkIpamANY
instead of network specific ipam_strategy, as it's
not relevant for subnet-level operations.

Fixes two problems with the /ip_addresses endpoint:
- 409 on POST with specified IP adresss for networks
with QuarkIpamBOTHREQ strategy
- Orphaned IPv6 or IPv4 (depending on requested version)
on POST to network with QuarkIpamBOTHREQ strategy


asadoughi:
LGTM

PR #305, opened 2015-01-12
Replaced use of xenstore_data with other_config

asadoughi:
RM11057


insequent:
-61 line code change, nice.

insequent:
Did something change recently, or is this just an optimization?

insequent:
With the caveat that I still don't know much about xenapi, this LGTM. :snowman:

asadoughi:
Yes, we're changing the internals as a result of discussions with the nova
team to maintain deployment independence and avoid any kind of collision.

insequent:
Working with the VIFs directly does look cleaner.

PR #306, opened 2015-01-13
redis_sg_tool now installs + fixes

Cerberus98:
RM11081

Moves redis_sg_tool to the quark namespace and adds an entry point to
setup.cfg for it. Additionally removes a group_by from
ports_with_security_groups_count db_api call that was breaking the
ports-with-groups command line action.

### DIFF
--- a/quark/security_groups/redis_sg_tool.py
+++ b/quark/security_groups/redis_sg_tool.py
@@ -97,8 +97,8 @@ class QuarkRedisTool(object):
         elif command == "write-groups":
             self.write_groups(self._dryrun)
         else:
-            print ("Redis security groups tool. Re-run with -h/--help for "
-                   "options")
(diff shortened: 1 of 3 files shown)


quark/security_groups/redis_sg_tool.py
```diff
@@ -258,8 +258,12 @@ def write_groups(self, dryrun=False):
...
-if __name__ == "__main__":
+def main():
     arguments = docopt.docopt(__doc__,
                               version="Quark Redis CLI %.2f" % VERSION)
     redis_tool = QuarkRedisTool(arguments)
     redis_tool.dispatch()
+
+
+if __name__ == "__main__":
+    main()
```
Cerberus98: Left this in so I didn't have to have quark actually installed to run it

Cerberus98:
WAT

mpath:
LGTM

PR #307, opened 2015-01-14
Replaced use of post_live_migrate a new quark xenapi plugin

asadoughi:
Review/Merge #305 first, please.

### DIFF
--- a/quark/agent/plugins/xenapi/etc/xapi.d/plugins/neutron_vif_flow
+++ b/quark/agent/plugins/xenapi/etc/xapi.d/plugins/neutron_vif_flow
@@ -1,0 +1,44 @@
+#!/usr/bin/python
+
+import subprocess
+
+import XenAPIPlugin
+
+script_command = ('/etc/xensource/scripts/vif_flow online vif-%s-%s -d')
+
+
+def check_output(*popenargs, **kwargs):
+    kwargs['close_fds'] = kwargs.get('close_fds', True)
+    input = kwargs.pop('input', None)
+
+    if input is not None and kwargs.get('stdin') != subprocess.PIPE:
+        kwargs['stdin'] = subprocess.PIPE
+
+    process = subprocess.Popen(stdout=subprocess.PIPE, *popenargs, **kwargs)
+    output, unused_err = process.communicate(input=input)
+    retcode = process.poll()
+
+    return (retcode, output)
+
+
+def get_output(*command, **kwargs):
+    return check_output(command, **kwargs)[1].strip()
+
+
+def get_return_code(*command, **kwargs):
+    return check_output(command, **kwargs)[0]
+
+
+def online_instance_flows(session, arg_dict):
+    domid = arg_dict['dom_id']
+    vif_index = arg_dict['vif_index']
+
+    command = script_command % (domid, vif_index)
+    retcode = get_return_code(*command.split(' '))
+    if retcode:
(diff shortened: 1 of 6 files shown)


asadoughi:
I'd like to manually test building the rpm and invoking the plugin before removing the `WIP`.

asadoughi:
I was able to build the RPM on Fedora VM in the cloud, but when I copied it over to XenServer and tried to install it I got a `cpio: Bad magic` error. So, for testing for now, I've manually copied over the neutron_vif_flow to /etc/xapi.d/plugins, ran chmod a+x, and invoked it from a XenAPI session from the compute node. I learned that plugins can only take in strings and must return strings out. I've updated the PR as such.

asadoughi:
Removed WIP from title. Please review.

quark/agent/tests/test_xapi.py
```diff
@@ -43,17 +41,26 @@ def test_get_instances(self):
...
+                        "domid": "1"},
             "opaque2": {"other_config": {},
                         "power_state": "", "is_a_template": True,
                         "is_control_domain": True,
-                        "name_label": ""},
+                        "name_label": "",
+                        "VIFs": [],
+                        "domid": "2"},
         }
         instances = self.xclient.get_instances()
```
insequent: So, I kinda wish that this was more clear about only showing power_state=running instances. Just a side note though, nothing to do with this PR.

quark/agent/plugins/xenapi/etc/xapi.d/plugins/neutron_vif_flow
```diff
@@ -0,0 +1,44 @@
...
+
+
+def online_instance_flows(session, arg_dict):
+    domid = arg_dict['dom_id']
+    vif_index = arg_dict['vif_index']
+
+    command = script_command % (domid, vif_index)
+    retcode = get_return_code(*command.split(' '))
+    if retcode:
+        raise Exception('online_instance_flows command failed |%s|' % command)
```
insequent: Might want to grab stderr no? Or even combine stdout and stderr, and return that?

insequent:
LGTM, besides little notes above.

Since the rpm is literally a file drop (i.e. no real compiling), fixing the rpmbuild for XenStore should be relatively easy, no? Want to hash that out first?

quark/agent/plugins/xenapi/etc/xapi.d/plugins/neutron_vif_flow
asadoughi: This script, I copied from an existing nova plugin so I didn't think about it much. I'm not sure what the limits are on the length of the string returned from here.

asadoughi:
I think the RPM is fine. I'm just not sure what went wrong between Fedora VM and trying to install it in XenServer. I think I'd like to get this merged and debug it once everything is integrated if it fails, jenkins will be creating the RPM .

quark/agent/plugins/xenapi/etc/xapi.d/plugins/neutron_vif_flow
andyhky: Just the return code would be plenty in this case. The stderr/stdout is already going to be logged locally on the hypervisor.

insequent:
LGTM :snowman:

jmeridth:
:+1:

PR #309, opened 2015-01-15
Delete MAC addresses in ranges marked do_not_use

Cerberus98:
RM11043

Implements use of the do_not_use column on mac_address_ranges.
Additionally, sets a rule in place that if a MAC is deallocated in a
range marked as do_not_use, it's deleted instead of deallocated. The
reason for this is two fold:

* We needed a fast fix to unblock other work being done with Neutron
* Introducing the logic necessary to handle deallocating and not reusing
  MACs in the current implementation would also introduce a new chance
  for deadlocks as MAC address ranges would have to be included in all
  queries for MAC address reallocation.
* I have an in-progress fix for removing the SELECT FOR UPDATE when
  reallocating MAC addresses, but there are flaws I have yet to work
  around. Rather than rush the fix, I decided to create a less than
  optimal path so we could move the ball forward.

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -325,6 +325,10 @@ def mac_address_find(context, lock_mode=False, **filters):
     return query.filter(*model_filters)
 
 
+def mac_address_delete(context, mac_address):
+    context.session.delete(mac_address)
+
+
 def mac_address_range_find_allocation_counts(context, address=None):
     count = sql_func.count(models.MacAddress.address)
     query = context.session.query(models.MacAddressRange,
@@ -336,6 +340,7 @@ def mac_address_range_find_allocation_counts(context, address=None):
         query = query.filter(models.MacAddressRange.last_address >= address)
         query = query.filter(models.MacAddressRange.first_address <= address)
     query = query.filter(models.MacAddressRange.next_auto_assign_mac != -1)
+    query = query.filter(models.MacAddressRange.do_not_use == False)  # noqa
     query = query.limit(1)
     return query.first()
 
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -113,6 +113,18 @@ def generate_v6(mac, port_id, cidr):
 
 
 class QuarkIpam(object):
+    def _delete_mac_if_do_not_use(self, context, mac):
+        # NOTE(mdietz): This is a HACK. Please see RM11043 for details
+        admin_ctx = context.elevated()
+        mac_range = db_api.mac_address_range_find(
+            admin_ctx, id=mac["mac_address_range_id"], scope=db_api.ONE)
+
+        if mac_range and mac_range["do_not_use"]:
+            # NOTE(mdietz): This is a HACK. Please see RM11043 for details
+            db_api.mac_address_delete(context, mac)
+            return True
+        return False
+
     @synchronized(named("allocate_mac_address"))
     def allocate_mac_address(self, context, net_id, port_id, reuse_after,
                              mac_address=None):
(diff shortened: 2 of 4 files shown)


Cerberus98:
I imagine this will be contentious, but let's chat about it

asadoughi:
Have you considered mac addresses that are already deallocated when a mac address range is marked as do_not_use? It seems like they will be able to be allocated.

Cerberus98:
That's what https://github.com/rackerlabs/quark/pull/309/files#diff-acddac4b6401f9da48acc821f21c024cR663 is for

The idea is people are going to migrate MAC address ranges we don't want into quark in the next few weeks, and we don't want anything new getting allocated in these ranges outside of their initial migration. As those ports are deleted over time, I want those MAC addresses tossed out rather than be potentially reallocatable.

I admit this is not a great solution (it's overloading what "do_not_use" really means versus subnets) but I'm told this is blocking and I don't have my solution for NOT locking MAC addresses on reallocation yet.

asadoughi:
The code you linked won't get invoked if already deallocated when the mac_address_range is get marked as full. I suggest either (1) deleting all deallocated MACs when the mac_address_range is marked as full or (2) doing a join on mac_address_range, filtering do_not_use=False when allocating a deallocated MAC.

I don't think migrating already deallocated MACs is part of the use case you're talking about, but I'd like to handle all cases, just in case.

Cerberus98:
do_not_use is not the same as full. Is full what you really mean? We treat it the same as subnets, where the indicator is -1

Already deallocated MACs shouldn't be migrated, if I understand their requirements correctly. They intend to create Ports for each VIF they need to migrate in, and have it retain the same MAC it had in a previous life. It's not a 1:1 migration of data, but rather the bits that are actually in use.

Also to point 2, as referenced in the ticket and my second bullet point, that would currently introduce a new potential deadlock since we'd be locking the mac address ranges table AND the mac addresses table now. Since we haven't solved that yet (again, I have something in progress but have yet to finish it), I don't think introducing a greater chance for failure is worth it for this.

asadoughi:
I edited the previous post after posting it, please re-read.

asadoughi:
OK, so doing a join is bad because of deadlocks, but let's go ahead with at least (1) or (3). I feel this feature could be abused in the future and then we'll end up with a bunch of already deallocated MACs being allocated in a mac_address_range that we've marked as do_not_use=True.

Cerberus98:
Gotcha

Which one is (3)?

I'm fine introducing code to delete the deallocated MACs if the range is marked as full, with the caveat that we have yet to mark a mac range as full anywhere since our strategy has been to create /24s everywhere, thus providing 17 million MACs everywhere we create them.

Also, hopefully we can fix the deadlock around MAC reallocation shortly and treat do_not_use the right way, meaning we can lose the bit about deleting a MAC in a do_not_use range in the next couple months.

asadoughi:
Did you refresh the page to re-read my edited post? Please ignore all uses of the word 'full'.

To reiterate, I'm suggesting adding the following to the currently proposed implementation:
(1) deleting all deallocated MACs when the mac_address_range is marked as do_not_use (on mac_address_range_update)
OR
(3) not allocating a deallocated MAC as returned from mac_address_find if the associated mac_address_range has do_not_use=True.

Cerberus98:
Bah, it didn't update the first time

asadoughi:
Yeah, I should stop trying to use the edit feature. It's convenient for only the poster.

Cerberus98:
Alright so:

1) The problem here is there's currently no programmatic way to toggle that flag. People have been going into the database. To my knowledge there's no way to do a CASCADE off the value of a column, leaving us with expectations to change as well. I'm fine with this otherwise, save for the caveat presented
3) The problem here is if there ARE deallocated, do_not_use macs in the DB, they'll eat retries since nothing will clean them up. We can delete them as we find them, which might be the safest path, but is going to require a fair amount more code to accomplish.

Cerberus98:
Also I was wrong, they're /29s, but that's still 500k addresses. Tangential at this point, I know. Interesting nonetheless

asadoughi:
Ah, ok. For (1), yeah, we need an admin API for do_not_use. I'd like to go ahead with (3). What makes it a 'fair amount' of code? In my head, it's only like another 5 lines and 2 more UT.

Cerberus98:
I think that's a reasonable estimate, actually. I was thinking of something else when I said that, but right now that works.

mpath:
:+1:

asadoughi:
LGTM

PR #315, opened 2015-01-21
Adds support for ICMP types and codes

Cerberus98:
RM11165

### DIFF
--- a/quark/protocols.py
+++ b/quark/protocols.py
@@ -31,27 +31,21 @@ ETHERTYPES = {
     "IPv6": 0x86DD
 }
 
+PROTOCOLS = {"icmp": 1, "tcp": 6, "udp": 17}
+
 # Neutron only officially supports TCP, ICMP and UDP,
 # with ethertypes IPv4 and IPv6
-PROTOCOLS = {
-    ETHERTYPES["IPv4"]: {
-        "icmp": 1,
-        "tcp": 6,
-        "udp": 17,
-    },
-    ETHERTYPES["IPv6"]: {
-        "icmp": 1,
-        "tcp": 6,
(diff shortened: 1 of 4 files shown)


quark/protocols.py
```diff
@@ -31,27 +31,29 @@
...
+        "udp": PROTOCOLS["udp"],
     },
     ETHERTYPES["IPv6"]: {
-        "icmp": 1,
-        "tcp": 6,
-        "udp": 17
+        "icmp": PROTOCOLS["icmp"],
+        "tcp": PROTOCOLS["tcp"],
+        "udp": PROTOCOLS["udp"]
     }
```
asadoughi: Seems redundant. Why can't the value just be `PROTOCOLS`?

Cerberus98: Good point. I knew that looked stupid. I guess I was thinking later if we add protocols to one but not the other (There are loads that are exclusive to one or the other)

quark/protocols.py
```diff
@@ -115,35 +109,54 @@ def validate_remote_ip_prefix(ethertype, prefix):
...
                                        port_range_max):
     if protocol in ALLOWED_WITH_RANGE:
-        # TODO(anyone): what exactly is a TCP or UDP rule without ports?
-        if (port_range_min is None) != (port_range_max is None):
-            raise exceptions.InvalidInput(
-                error_message="For TCP/UDP rules, port_range_min and"
-                              "port_range_max must either both be supplied, "
-                              "or neither of them")
+        if protocol == PROTOCOLS["icmp"]:
+            if port_range_min is None and port_range_max is not None:
```
tr3buchet: type shouldn't be required

PR #316, opened 2015-01-21
Update security groups redis client to use security groups specific hash

TMaddox:
Actually expecting hash attr to be "security group rules" not "sg"


Cerberus98:
:+1: Already ran tests on my machine from this branch.

asadoughi:
:+1: we should be more careful about schema changes once we're in prod. We're having to drop all the data in netdev to test the latest code.

TMaddox:
@asadoughi  :+1: x :100:

Cerberus98:
Screw it, I saw the tests pass on my machine

PR #317, opened 2015-01-23
Refactoring Redis ClientBase out of security groups

TMaddox:
To make room for other uses of the Redis cache, it
is necessary to refactor the core components of a
ClientBase out of the existing Security Groups
Redis client. Afterwards I created a child
SecurityGroupsClient to implement the Security
Groups specific features.


quark/cache/security_groups_client.py
```diff
@@ -0,0 +1,170 @@
...
+                    destination = netaddr.IPNetwork(
+                        rule["remote_ip_prefix"])
+                    destination = str(destination.ipv6())
+
+            optional_fields = {}
+
+            # NOTE(mdietz): this will expand as we add more protocols
+            if rule["protocol"] == protocols.PROTOCOLS["icmp"]:
+                optional_fields["icmp type"] = rule["port_range_min"]
+                optional_fields["icmp code"] = rule["port_range_max"]
```
TMaddox: This is from: https://github.com/rackerlabs/quark/commit/cb5a019837869d9db828262dfc47b0f643f46d2f

Should I go ahead and add underscores?

quark/cache/security_groups_client.py
```diff
@@ -0,0 +1,170 @@
...
+            else:
+                optional_fields["port start"] = rule["port_range_min"]
+                optional_fields["port end"] = rule["port_range_max"]
```
TMaddox: Same as above

TMaddox:
Looks like an old .tox?

TMaddox:
Apparently I'm a baddie and forgot to flake8 test the unmanaged.py/test_unmanaged.py changes. Whoops!

quark/cache/redis_base.py
```diff
@@ -0,0 +1,181 @@
...
+        return redis.StrictRedis(**kwargs)
+
+    def vif_key(self, device_id, mac_address):
+        mac = str(netaddr.EUI(mac_address))
+
+        # Lower cases and strips hyphens from the mac
+        mac = mac.translate(MAC_TRANS_TABLE, ":-")
+        return "{0}.{1}".format(device_id, mac)
+
+    def _handle_connection_op(self, op, args=(), target=None):
```
asadoughi: Hmm, interface is odd, but I guess that's what you have to do to abstract except handling in one place.

asadoughi:
Nothing stands out to me at the moment, but I haven't slept much so I'll take a look again on Monday. Only action item I'd ask is if you could squash your 3 commits into one, please. Thanks.

quark/cache/redis_base.py
TMaddox: Hehe, agreed. I'm definitely open for suggestions. I mean, I could use a decorator too, if that's cleaner. Thinking about it, it actually seems cleaner to me, since really this is just a  rudimentary decorator now. :)

asadoughi: Yeah, a decorator would be a suitable pattern, here.

quark/cache/redis_base.py
```diff
@@ -0,0 +1,184 @@
...
+    @handle_connection_error
+    def vif_keys(self, field=None):
+        keys = self._client.keys("*.????????????")
+        filtered = []
+        if isinstance(keys, str):
+            keys = [keys]
+        for key in keys:
+            cmd = ('hget', (key, field)) if field else ('hgetall', (key,))
+            if getattr(cmd[0], self._client)(*cmd[1]):
+                filtered.append(key)
```
asadoughi: I think you've outclevered yourself here; the ordering of getattr args is wrong. Consider simplifying:

```python
value = None
if field:
   value = self._client.hget(key, field)
else:
   value = self._client.hgetall(key)
if value:
   filtered.append(key)
```

asadoughi: Also, failure of ordering, means there's not UT coverage here so I'd like more UT for this. Thanks.

TMaddox: Haha, spot on. Yeah, I'll simplify and add coverage. Thanks for catching that.

quark/cache/security_groups_client.py
Cerberus98: Did the flows side change to include underscores? I was following the original schema, where underscores were omitted from multi-word keys like "port range min"

quark/cache/security_groups_client.py
Cerberus98: Ditto :-)

Cerberus98:
Only thing I'm curious about is your schema questions. Otherwise it's totes :+1: :eggplant: :hankey:

quark/cache/security_groups_client.py
TMaddox: Sorry, didn't realize that was the relationship. You may all ignore this and walk on by.

quark/cache/security_groups_client.py
TMaddox: Ditto the 2nd

PR #318, opened 2015-01-27
Traceback + refactoring

Cerberus98:
RM11181

Updates the quark agent code to use the path for the security groups
code, as well as fixes a bug where the agent always assumes it knows the
most recent state of xenstore. Finally, moves the tests/ folder from
under quark/agent to where the belong.


quark/agent/xapi.py
```diff
@@ -142,19 +142,36 @@ def get_interfaces(self, instances):
...
+                if not self.does_record_exist(session, vif.ref):
+                    session.xenapi.VIF.add_to_other_config(
+                        vif.ref, self.SECURITY_GROUPS_KEY,
+                        self.SECURITY_GROUPS_VALUE)
+            except XenAPI.Failure as xfe:
+                # We shouldn't lose all of them because one failed
+                # An example of a continuable failure is the VIF was deleted
+                # in the (albeit very small) window between the initial fetch
+                # and here.
+                LOG.exception(xfe)
```
asadoughi: Minor nit: I don't think passing the exception to `LOG.exception` is the intended use case.

Cerberus98: I felt the same, but then I got an argument exception. I'm wondering if they changed the format of the exceptions in Neutron. That or it was late and I was being stupid

Cerberus98:
btw, all this code was tested in netdev, including the respin.

quark/agent/xapi.py
```diff
@@ -142,19 +142,37 @@ def get_interfaces(self, instances):
             interfaces.add(VIF(device_id, rec["MAC"], vif_ref))
         return interfaces
 
+    def does_record_exist(self, session, vif_ref):
+        vif_rec = session.xenapi.VIF.get_record(vif_ref)
+        if vif_rec["other_config"].get("security_groups"):
```
asadoughi: Missed this earlier, sorry. `self.SECURITY_GROUPS_KEY` in place of `"security_groups"`

PR #319, opened 2015-01-29
Makes the agent apply flows to the correct VIF

Cerberus98:
RM11249

Previously, the agent assumed that the order of VIFs returned from Xen
was significant, and used the index in the list to called the flows
plugin. However, that isn't always strictly true. The right answer is to
get the device number from the VIF record itself and pass that into the
call.

### DIFF
--- a/quark/agent/xapi.py
+++ b/quark/agent/xapi.py
@@ -26,6 +26,7 @@ CONF = cfg.CONF
 agent_opts = [
     cfg.StrOpt("xapi_connection_url"),
     cfg.StrOpt("xapi_connection_username", default="root"),
+    cfg.IntOpt("xapi_enable_groups_retries", default=5),
     cfg.StrOpt("xapi_connection_password")
 ]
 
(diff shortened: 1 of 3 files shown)


quark/agent/xapi.py
```diff
@@ -180,18 +190,17 @@ def _unset_security_groups(self, session, interfaces):
...
-                vif_ref_to_vm[vif_ref] = i, vm
-
         for vif in interfaces:
             try:
-                vif_index, vm_vif = vif_ref_to_vm[vif.ref]
-            except KeyError:
-                LOG.info("vif ref %s (%r) not found on any VM", vif.ref, vif)
+                vif_rec = session.xenapi.VIF.get_record(vif.ref)
+                vm_rec = session.xenapi.VM.get_record(vif_rec["VM"])
+                vif_index = vif_rec["device"]
```
asadoughi: I went to some lengths to separate data ingestion from application, but perhaps it was a wasteful effort. You should be able to remove `interfaces` as an argument if you'd like to continue down that path.

quark/agent/xapi.py
```diff
@@ -150,15 +153,22 @@ def does_record_exist(self, session, vif_ref):
...
 
+    @utils.retry_loop(CONF.AGENT.xapi_enable_groups_retries, delay=0,
+                      backoff=2)
+    def _apply_and_wait_for_security_groups(self, session, vif):
+        if not self.does_record_exist(session, vif.ref):
+            session.xenapi.VIF.add_to_other_config(
+                vif.ref, self.SECURITY_GROUPS_KEY,
+                self.SECURITY_GROUPS_VALUE)
+            if not self.does_record_exist(session, vif.ref):
+                raise Exception("VIF tag not yet applied!")
```
asadoughi: Can you explain this a bit?

asadoughi:
Good bug sleuthing. I should have known better that the index of the `vm["vif"]` list is not related to the device index. Yipes.

quark/agent/xapi.py
Cerberus98: yeah, I should probably fetch all this once at the top and pass it down. I was just lazy and trying to get it done since it's "blocking" :-/

quark/agent/xapi.py
Cerberus98: I'm debating whether or not we need it. When I was troubleshooting this yesterday, Koelker and Trey thought that rules weren't being applied because of a race condition where the tag wasn't set before the plugin was called. When this didn't fix it, I moved on to the VIF ordering dealio. I was considering leaving it in in case it's a problem, but since I can't prove it helps anything, I'm gonna remove it.

Cerberus98:
Tell you what. I'm going to remove the retry_loop and resubmit as is. I'd like to get this deployed functional so testing can continue, and I'll work on a refactoring patch for _refresh_interfaces this afternoon. Thoughts?

asadoughi:
As you wish.

PR #320, opened 2015-01-30
Revert "eager-loading of associations relationship on IPAddress model"

TMaddox:
This reverts commit 494a3a4ddc1d7819e761df5eb0bc080b0c877337.

Determined that lazy='subquery' on the associations attribute of the
IPAddress model was substantially increasing the volume of
DBDeadLock errors and thusly degrading API performance. This is to
revert that model change. There were many reverts of this before for
producing test packages. This is intended to be the final removal of
this commit.

RM11222


jmeridth:
:+1:

PR #321, opened 2015-02-02
Reverts RM10879

Cerberus98:
This is the patch that fixes the TOCTOU bug. However, I apparently
failed to test what happens with v6 subnets. The problem is MySQL has no
native 128 bit type. Rather than trying to patch it as is, I decided to
revert this patch and revisit as a whole.


jmeridth:
:+1:

roaet:
:hatched_chick:

PR #323, opened 2015-02-05
Fixes multiple Exception signatures

Cerberus98:
RM11250

Updates the security groups plugin to properly instantiate the
exceptions being raised. The bug was two fold:

When testing with neutronclient, it always issues a GET
/<resource>?id=<uuid> on the resource, returning an empty list if
the security group or rule in question doesn't exist. In those cases,
the DELETE or GET as appropriate were never called on the routes where
the exceptions were being raised incorrectly.

Secondly, tests passed because of save_and_reraise in Neutron, which
would reraise our SecurityGroupNotFound and SecurityGroupRuleNotFound
exceptions, allowing the tests looking for those to pass. However, the
message included with those exceptions was the traceback of the
respective exceptions failing to instantiate, rather than the expected
message.

### DIFF
--- a/quark/plugin_modules/security_groups.py
+++ b/quark/plugin_modules/security_groups.py
@@ -91,7 +91,7 @@ def create_security_group_rule(context, security_group_rule):
         group = db_api.security_group_find(context, id=group_id,
                                            scope=db_api.ONE)
         if not group:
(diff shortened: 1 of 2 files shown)


asadoughi:
:+1:

jmeridth:
oops, will wait on CI next time.  :sunglasses:

PR #324, opened 2015-02-07
Changes use of messaging to oslo_messaging

Cerberus98:
An upstream neutron change moved neutron.common.rpc.messaging to
neutron.common.rpc.oslo_messaging, breaking several of our unit test
modules. This patch changes the mocks to compensate.

PR #325, opened 2015-02-07
Rm11364

Cerberus98:


jmeridth:
Why does this one and #326 have two of the same commits?

Cerberus98:
Because I rebased them so they could merge in series. They would all fail tests without #324 but I didn't want to make it part of the same patch. Additionally, I need the two latter patches in whatever branch we're planning to deploy, but I don't think our pinned version of neutron is far enough along yet that we want the oslo_messaging fix there.

Cerberus98:
Though now the first one merged and it doesn't look to be referencing that :-/

jmeridth:
@tr3buchet review this please

quark/cache/security_groups_client.py
```diff
@@ -30,8 +30,20 @@
...
 
+ALL_V4 = netaddr.IPNetwork("::ffff:0.0.0.0/96")
+ALL_V6 = netaddr.IPNetwork("::/0")
+
 
 class SecurityGroupsClient(redis_base.ClientBase):
+    def _convert_remote_network(self, remote_ip_prefix):
+        # NOTE(mdietz): RM11364 - While a /0 is valid and should be supported,
+        #               it breaks OVS to apply a /0 as the source or
+        #               destination network.
```
asadoughi: This strikes me that we're solving this at the wrong layer. i.e. DP code should translate this before applying it.

tr3buchet: The API should just all through whatever, and we should figure it out on the backend you mean?

asadoughi: You mean pass through? I don't think pass through is the right mindset, but I would towards aiming for consistency in terms of the data between the control plane and data plane. There might be constraints I'm not aware of as well... .

Cerberus98: Honestly I just got to it first. It may make more sense to put it in the flow scripts, but I thought about how easy our deploys are versus flows, coupled with me just trying not to break my callee with bad data

tr3buchet: API should sanitize input, no?

Cerberus98:
Long story short, I don't care. I'd kind of prefer to just have it done since I have a clear way of both fixing it and unit testing it, whereas I imagine that's not the case with the flows scripts. However, I can be sold on the argument that it belongs elsewhere.

Cerberus98:
Which way is it going to be? Otherwise I'm just gonna merge it and get on with my day

quark/cache/security_groups_client.py
asadoughi: @Cerberus98 that's acceptable as long as we can all agree that this is a HACK because the data plane cannot handle this input.

asadoughi:
I'm ok with this as long as we acknowledge this is a HACK because the data plane cannot handle this input. i.e. quark is incurring technical debt because of data plane limitations.

tr3buchet:
This is not a hack. It's API sanitization. Allowing traffic in from a /0 is the same thing as not specifying source network. OVS doesn't support /0 because of this fact.

asadoughi:
Smells like a duck, walks like a duck, quacks like a duck. It's a hack.

PR #326, opened 2015-02-07
Rm11363

Cerberus98:


quark/cache/redis_base.py
```diff
@@ -181,6 +181,10 @@ def delete_field(self, key, field):
         return self._client.hdel(key, field)
 
     @handle_connection_error
+    def delete_key(self, key):
```
insequent: Super minor, but the order of these methods seem a bit arbitrary.

jmeridth: @insequent huh?

jmeridth:
:+1: :cookie: :snowboarder:

quark/cache/redis_base.py
Cerberus98: I'll second that huh. Are you saying that delete_key should be before delete_field?

quark/drivers/unmanaged.py
```diff
@@ -77,6 +77,12 @@ def update_port(self, context, port_id, **kwargs):
     def delete_port(self, context, port_id, **kwargs):
         LOG.info("delete_port %s %s" % (context.tenant_id, port_id))
 
+        # Contacting redis is cheaper than hitting the database to find out
+        # if we have rules to delete, and deleting an absence of rules is a
+        # NOOP, so this is a safe operation
+        client = sg_client.SecurityGroupsClient(use_master=True)
+        client.delete_vif(kwargs["device_id"], kwargs["mac_address"])
```
insequent: Another cosmetic question, why is kwargs so heavily used in here? Why not explicitly call `device_id` and `mac_address`?

quark/cache/redis_base.py
insequent: Your methods are not in alphabetical or lumped by purpose. Talking with @jmeridth, I don't think there's a standard anywhere, but I guess I'm just a creature of order. You have:

get_field
delete_field
delete_key
get_fields

Why not group gets and deletes together, or even alphabetical order. If I'm looking through someone else's code, and don't know what to expect, I like to have thing in an order that makes it easy to find things and ensures I'm not missing something.

quark/drivers/unmanaged.py
Cerberus98: Lack of foresight into what parameters would be needed to satisfy these requests (Security Groups™ the feature designated these as the necessary keys long after this interface was created) coupled with the need for a larger refactor we have yet to undertake.

quark/cache/redis_base.py
insequent: Like I was saying though, super minor, and just a preference on my side. Definitely not a requirement.

quark/drivers/unmanaged.py
insequent: Cool, cool. Works either way.

quark/cache/redis_base.py
Cerberus98: Yeah, I get you. Just wasn't clear what you meant, since refactoring wasn't the intent of this patch. Would you mind making couple of redmines for general cleanup and refactoring? That way we can separate the implementation/bug fix needed here from a larger redo

insequent:
:snowman:

Cerberus98:
Going to assume thumb + snowman == merge

PR #327, opened 2015-02-09
Re-points the entrypoint for redis_sg_tool

Cerberus98:
RM11375

A recent refactoring moved redis_sg_tool and forgot to give it an
__init__.py to make it properly import and also forgot to ticket the
entrypoint.


roaet:
LGTM

PR #330, opened 2015-02-10
_unset now catches Xen failures and continues

Cerberus98:
RM11399

XenServer, in it's infinitely inconsistent wisdom, allows one to remove
non-existent parameters from known objects without raising, but does not
allow one to remove non-existent parameters from non-existent objects
without raising.

A failure caused by bad data was going unhandled in testing, causing us
to never apply flows on boxes where the _unset step was blowing up.


asadoughi:
:+1: pending tests

PR #332, opened 2015-02-11
Fixes building the tox venv after upstream change

Cerberus98:
Upstream change https://github.com/openstack/neutron/commit/dea4755
supplied a new environment variable NO_EVENTLET that causes all sorts of
shenanigans when building a venv including neutron. Adding it to quark's
tox.ini fixes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 python:
   - "2.7"
   - "2.6"
+before_install:
+  - "export NO_EVENTLET=1"
 install:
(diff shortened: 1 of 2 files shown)


Cerberus98:
Ok honestly travis, wtf. Including that fixed it for me

Cerberus98:
I'm thinking travis doesn't include the tox.ini from the patch being built. Removing that lines causes the same error as travis is reporting. Adding it in fixes it on my side. Wondering if we need to push this into the repo regardless

Cerberus98:
Ahh, forgot about the travis.yml

PR #333, opened 2015-02-12
GET /security-groups now responds with rules

Cerberus98:
RM11427

Updates the views for security groups to always include the rules rather
than lists of rule ids.


asadoughi:
:+1: pending tests

PR #336, opened 2015-02-17
Replace tenant_id with segment_id

asadoughi:
segment_id is the correct column to use in ip_availability because
segment_id is what is used on the nova side when allocating ports to
indicate the cell.

RM11394


jmeridth:
:+1:

PR #338, opened 2015-02-17
Replaces agent JSON control with Redis control

Cerberus98:
RM11449

This patch rips out the JSON version control system implemented in a
previous version of the agent and replaces it with a new hash field in
Redis. The agent had to be substantially reworked to compensate for the
change. Additionally, this patch enables duplex communication for a
process using the quark redis client by implementing discreet read and
write only connection pools.


Cerberus98:
This is a huge change and may require some cleanup yet. I wanted to get it out sooner rather than later, however, so fire away

Cerberus98:
I'll be updating this with a patch to randomly pick a slave connection ala the flows scripts.

quark/agent/xapi.py
```diff
@@ -124,15 +126,14 @@ def get_instances(self):
...
 
-    def get_interfaces(self, instances):
+    def get_interfaces(self):
         """Returns a set of VIFs from `get_instances` return value."""
         LOG.debug("Getting interfaces from Xapi")
 
         session = self._session()
-        try:
+        instances = self.get_instances(session)
+        with self.sessioned() as session:
```
Cerberus98: forgot to clean this up

quark/cache/security_groups_client.py
```diff
@@ -145,40 +146,55 @@ def apply_rules(self, device_id, mac_address, rules):
         if not self._use_master:
             raise q_exc.RedisSlaveWritesForbidden()
 
-        ruleset_uuid = str(uuid.uuid4())
-        rule_dict = {SECURITY_GROUP_VERSION_UUID_KEY: ruleset_uuid,
-                     SECURITY_GROUP_RULE_KEY: rules}
+        rule_dict = {SECURITY_GROUP_RULE_KEY: rules}
         redis_key = self.vif_key(device_id, mac_address)
         self.set_field(redis_key, SECURITY_GROUP_HASH_ATTR, rule_dict)
+        self.set_field_raw(redis_key, SECURITY_GROUP_ACK, False)
```
Cerberus98: Will pipeline this in a future patch

quark/cache/security_groups_client.py
```diff
@@ -145,40 +146,55 @@ def apply_rules(self, device_id, mac_address, rules):
...
 
     @utils.retry_loop(3)
-    def get_security_groups(self, new_interfaces):
+    def get_security_group_states(self, interfaces):
         """Gets security groups for interfaces from Redis
 
-        Returns a dictionary of xapi.VIFs mapped to security group version
-        UUIDs from a set of xapi.VIF.
+        Returns a dictionary of xapi.VIFs mapped to security group ack'd
+        booleans from a set of xapi.VIF.
```
Cerberus98: This comment looks stupid

quark/agent/agent.py
```diff
@@ -45,51 +44,86 @@ def _sleep():
...
 
 
+def partition_vifs(xapi_client, interfaces, security_group_states):
+    """Splits VIFs into three categories
+
+    Added - Groups exist in Redis that have not been ack'd and the VIF
+            is not tagged
+    Updated - Groups exist in Redis that have not been ack'd and the VIF
+              is already tagged
+    Removed - Groups do NOT exist in Redis but the VIF is tagged
```
tr3buchet: what about 
groups exist in redis that have been ack'd and the vif is not tagged? that 4th one we talked about

also, if you don't mind, can you add the response action to each of these? This seems as good a place as any to document

Cerberus98: It's actually there, in the comments right below where you marked this

quark/agent/xapi.py
```diff
@@ -218,10 +222,8 @@ def update_interfaces(self, instances, added_sg, updated_sg, removed_sg):
             return
 
         session = self._session()
```
asadoughi: Extra.

asadoughi:
There are a lot of things I could nit pick here, but instead I'll just note that the structure/maintainability of this code is going from bad to worse. I think it's fine to hack and slash for a first working prototype, but I expect a task for refactoring a lot of this before going to general availability.

Cerberus98:
I'd love to know what these nitpicks are. Considering the agent didn't work at all in the first place, the idea that it's shifted in maintainability is ridiculous.

tr3buchet:
@asadoughi is that a joke?

asadoughi:
I forgot to commend you on quickly turning around a new implementation: Awesome job.

One way I view maintainability is the way the data flows through a program. The way the data flows from xapi and Redis and back through xapi and Redis is not straightforward to me in the proposed implementation without a lot of mental notetaking. Another way to look at maintainability is how quickly would one be able to debug the program if there were a bug to present itself. 

A couple of nits I recall were interface definitions and exception messages. There are more if you'd like me to go ahead and enumerate them.

andyhky:
@asadoughi - 

We took a look at the earlier model and had issues with diagnosing if the agent has seen the change or not. It was a little cumbersome checking a UUID in redis, then on the filesystem and to see if they're different. We chose an "acked" field to be very explicit around if the quark-agent has processed a particular security group or not.

The schema change has been documented internally in Redmines but the intent behind them may not have been well captured.

asadoughi:
@andyhky That makes sense. I'm not debating the design since I wasn't present for the design discussion. I'm just talking about how the code looks here given the design intents, the implementation.

Cerberus98:
I'm trying to rationalize what *new* flow problems there are. Admittedly is_vif_tagged is a bad method, but it's also in a bad spot because we can't care too hard if it fails. The show much go on.

However, it's not a whole lot different from before. The main difference is our source of truth regarding which VIFs are actually tagged is different. The responsibility has shifted here to the agent, which is now acting more appropriately as the controller. Previously we leaked instances into the agent for no real gain. After we switched to using get_record to get the correct device_id for the VIF, instances was no longer needed outside of the xapi client.

I'm currently cleaning up what things I think I can in the patch with the time we have, but it needs to merge sooner rather than later since Miguel is calling us blocked until the JSON store is dead.

asadoughi:
There aren't new data flow problems, just new instances of them: writing to Redis and an additional read from xapi being amongst them. What is the current state of the system if `is_vif_tagged` fails but the `get_interface` call succeeds? What is the current state of the system if the Redis write `ack_groups` fails, but the write to xapi succeeds? These kinds of things are harder for me to understand when the data input/output is spread out.

Cerberus98:
The current state is: Anything that impedes the REST of the flows from being applied has to be ignored. I've fixed multiple errors over the last couple weeks because the agent didn't handle any of those cases. 

If the implementation here is *not* spread out the whole thing is tightly coupled. The agent now acts as the adapter layer between these two disparate systems, and ideally all things have to err on the side of applying what looks to be the correct state to the instance. Considering there's no mechanism currently to indicate any sort of failure in the system, it doesn't matter if we can't ack the groups. We have to carry on and provide the state we can see even if we can't write it back.

Frankly I'm not even sure how you can argue that it's confusing now when the agent had almost zero handling for bad state previously. Confusing relative to what?

I currently have refactored the use of the xapi layer such that we only fetch the records once per run, period, and thus removes the is_vif_tagged call entirely. I still need to get the tests passing, but it's a marked improvement over the use of extra calls and what other nonsense.

tr3buchet:
@asadoughi the refactor will make drastically fewer xapi calls and the same number of calls to redis 99+% of the time. The only extra calls to redis will be one write and will only happen in the rare event a security group update.

@asadoughi Let's not forget the whole reason the refactor is necessary - If we were replacing a perfectly adequate agent with a different agent for like no reason at all, then sure let's sit around and bike shed. But it's not perfectly adequate, it's the primary cause of failure in functional testing; it's 11th hour blocking the control plane SG feature release.

Say nothing or help. Snarky nitpicks or commendations are not welcome.

asadoughi:
@Cerberus98 Sure, not allowing one vif's failure to not block other vif's from failing makes sense. However, once that's established it adds complexity in error handling as you described. I think there's another error scenario that needs to be handled in that vein: The agent should not ack the groups that failed to apply to xapi, right? I await your latest patch.

@tr3buchet Cool. I don't want to bikeshed or further delay this PR. What I am trying to say is, and perhaps I'm saying it to the wrong audience, is that we are purposefully not creating the best version of the software we can because of time pressure and we will need to address that in the future. I think I am saying it to the wrong audience since we all already know that and I should be convincing management of that. I think we have a large enough backlog queue already.

tr3buchet:
@asadoughi management is already aware.

tr3buchet:
looks good to me so far!

quark/agent/xapi.py
jmeridth: Making a new PR for that line removal.  Running tests locally and will push.

PR #341, opened 2015-02-18
Added GET /ip_availabilities API

asadoughi:
Currently, only single value per parameter is able to be specified
due to limitations of the resource handler. For example, GET
/ip_availabilities?ip_version=4&ip_version=6 will only pass in the
filter for ip_version equal to 6.

RM7008


jmeridth:
:+1:

PR #342, opened 2015-02-19
Refactors RM11449

Cerberus98:
The agent now correctly caches and uses the VIFs the first time it pulls
them from XAPI, rather than fetching them over and over. Additionally
implements handling for ack'ing VIFs only if they actually tagged and
applied their flows.


tr3buchet:
LGTM

PR #343, opened 2015-02-19
Remove deallocated filter when given specific IP

roaet:
RM11361


Cerberus98:
:snowboarder: :+1: :egg: :hankey: :cn:

PR #344, opened 2015-02-19
Add ability for admin to allocate mac addresses from do_not_use ranges

sumanthns:
This adds ability for admin to allocate
mad address from a do not use range while
creating a port, by passing an arg in
request.

### DIFF
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -127,12 +127,14 @@ class QuarkIpam(object):
 
     @synchronized(named("allocate_mac_address"))
     def allocate_mac_address(self, context, net_id, port_id, reuse_after,
-                             mac_address=None):
+                             mac_address=None,
+                             use_forbidden_mac_range=False):
         if mac_address:
             mac_address = netaddr.EUI(mac_address).value
 
         kwargs = {"network_id": net_id, "port_id": port_id,
-                  "mac_address": mac_address}
+                  "mac_address": mac_address,
+                  "use_forbidden_mac_range": use_forbidden_mac_range}
         LOG.info(("Attempting to allocate a new MAC address "
                   "[{0}]").format(utils.pretty_kwargs(**kwargs)))
 
@@ -185,7 +187,9 @@ class QuarkIpam(object):
             with context.session.begin():
                 try:
                     fn = db_api.mac_address_range_find_allocation_counts
-                    mac_range = fn(context, address=mac_address)
+                    mac_range = \
+                        fn(context, address=mac_address,
+                           use_forbidden_mac_range=use_forbidden_mac_range)
 
                     if not mac_range:
                         LOG.info("No MAC ranges could be found given "
--- a/quark/plugin_modules/mac_address_ranges.py
+++ b/quark/plugin_modules/mac_address_ranges.py
@@ -91,11 +91,13 @@ def create_mac_address_range(context, mac_range):
         raise exceptions.NotAuthorized()
 
     cidr = mac_range["mac_address_range"]["cidr"]
+    do_not_use = mac_range["mac_address_range"].get("do_not_use", "0")
     cidr, first_address, last_address = _to_mac_range(cidr)
     with context.session.begin():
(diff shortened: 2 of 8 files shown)


quark/plugin_modules/ports.py
```diff
@@ -161,10 +164,12 @@ def _allocate_ips_undo(addr):
                         LOG.exception("Couldn't release IP %s" % address)
 
         @cmd_mgr.do
-        def _allocate_mac(net, port_id, mac_address):
+        def _allocate_mac(net, port_id, mac_address,
+                          forbidden_mac_range=False):
```
asadoughi: Nit: Variable name changed from `use_forbidden...` to `forbidden...`?

quark/plugin_modules/mac_address_ranges.py
```diff
@@ -91,11 +91,13 @@ def create_mac_address_range(context, mac_range):
...
 
     cidr = mac_range["mac_address_range"]["cidr"]
+    do_not_use = mac_range["mac_address_range"].get("do_not_use", "0")
     cidr, first_address, last_address = _to_mac_range(cidr)
     with context.session.begin():
         new_range = db_api.mac_address_range_create(
             context, cidr=cidr, first_address=first_address,
-            last_address=last_address, next_auto_assign_mac=first_address)
+            last_address=last_address, next_auto_assign_mac=first_address,
+            do_not_use=do_not_use)
```
asadoughi: Just thinking out loud: Previously, we didn't expose `do_not_use` for `mac_address_range`s or `subnet`s. Not exposing `subnet.do_not_use` makes sense, but `mac_address_range`s are not exposed to the customer so this change makes sense.

quark/ipam.py
```diff
@@ -185,7 +187,9 @@ def allocate_mac_address(self, context, net_id, port_id, reuse_after,
             with context.session.begin():
                 try:
                     fn = db_api.mac_address_range_find_allocation_counts
-                    mac_range = fn(context, address=mac_address)
+                    mac_range = \
+                        fn(context, address=mac_address,
+                           use_forbidden_mac_range=use_forbidden_mac_range)
```
asadoughi: Nit: just use the function name if it's going to be multi-line anyway. For example,
```
mac_range = db_api.mac_address_range_find_allocation_counts(
    ...)
```
Or, is that over the 79th column? If so, I'm sorry for such a long function name.

asadoughi:
Finished reviewing first draft. LGTM. I'm curious if anyone can recall the reasoning of why we didn't expose `mac_address_range.do_not_use` before I'm curious to hear it.

Cerberus98:
Previously mac_address_ranges weren't exposed because the column was implemented independently of the code to actually make it work, for no reason I'm aware of. Additionally, I simply forgot. When I provided the IPAM portion to honor the column, I didn't modify the plugin portion at all, partially because I knew the mac_ranges were being created by hand manually since it happens so infrequently (only when we add entirely new environments. I didn't think about the change in interface required to go along with it.

asadoughi:
@Cerberus98, makes sense. Would it make sense to add an update (PUT) method for `mac_address_range.do_not_use` as well?

Cerberus98:
Yeah, for posterity, I think it would.

quark/ipam.py
sumanthns: @asadoughi, the function name is over 79 column. So I am leaving it as it for now.

sumanthns:
@asadoughi, @Cerberus98 , pushed some changes based on comments. Can you take a look again?

asadoughi:
LGTM. We should file an issue for the update request, for work later as a separate PR.

sumanthns:
added https://github.com/rackerlabs/quark/issues/345

Cerberus98:
Looks ok to me. I was nervous about the "use_forbidden_mac_range" bit (also not the biggest fan of the name) but I'll go ahead and merge this

jkoelker:
@Cerberus98 yea `omg_do_use` would have been better ;)

PR #347, opened 2015-02-24
Adds support for ICMPv6

Cerberus98:
RM11567

Updates the protocol handling code to allow for different sets of
protocol integers between ethertypes. Additionally, adds the correct
protocol number for ICMP under IPv6


jmeridth:
:+1:

PR #349, opened 2015-03-03
TOCTOU 2 part 2 the deuce

Cerberus98:
RM10879

Reworks the original patch for RM10879 by completely stripping out
sequential IPv6 address generation, which never made sense anyway.
Instead, when generating a v6, for the edge cases where we would have
generated a sequential address previously, we now use the rfc3041 method
generating a random integer from the port_id within the subnet prefix.
As a result, we also no longer update next_auto_assign_ip on v6 subnets.

Technically this means we no longer need to lock rows when allocation v6
addresses, but I wanted to keep this patch as simple as possible.

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -484,9 +484,41 @@ def subnet_find_ordered_by_most_full(context, net_id, **filters):
         query = query.filter(models.Subnet.ip_version == filters["ip_version"])
     if "segment_id" in filters and filters["segment_id"]:
         query = query.filter(models.Subnet.segment_id == filters["segment_id"])
+    query = query.filter(models.Subnet.next_auto_assign_ip != -1)
+
     if "subnet_id" in filters and filters["subnet_id"]:
         query = query.filter(models.Subnet.id.in_(filters["subnet_id"]))
+    return query
+
+
+def subnet_update_next_auto_assign_ip(context, subnet):
+    query = context.session.query(models.Subnet)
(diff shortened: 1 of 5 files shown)


Cerberus98:
90% of this patch merged previously. The new significant changes are all around v6 addressing.

asadoughi:
Could you publish a diff of patches with the previously approved patch, please? It would make it easier to review. :)

jmeridth:
@asadoughi aren't those links in the redmine? #302 was the previous PR that got reverted.

asadoughi:
I guess I was looking for something more along the lines of `diff $(git diff -p) $(git diff -p)`. I'll post the result when I get around to it.

Cerberus98:
I'm not sure how to generate ONLY the diff you're asking for using git diff. The idea you provided above definitely doesn't accomplish it by itself, since treats the SHAs as a range, not as a diff in isolation. I generated left and right using git show instead, and the result was also not quite what you'd expect because it gets tripped up on line numbers. I'm sure there's a way to do it, but I already spent more time on it than I think is worthwhile.

You could have readily just reviewed the patch by now. It may be mostly previously approved code, but it only works in aggregate. If that's a problem, feel free to skip it. I can harass the others to do it.

asadoughi:
Yeah, I was kinda being lazy. I'll open two windows to see what I can ignore. I like gerrit for this exact feature.

asadoughi:
LGTM :8ball:

PR #350, opened 2015-03-06
MAC Range TOCTOU fix

Cerberus98:
RM11012

Updates Quark mac address ranges to atomically update
next_auto_assign_mac. This fix borrows heavily from the subnet atomic
update fix. Please see RM10879 for more details.

### DIFF
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -213,8 +213,7 @@ class QuarkIpam(object):
                         # knowing, so set the next_auto_assign to be -1
                         # so we never try to create new ones
                         # in this range
-                        rng["next_auto_assign_mac"] = -1
-                        context.session.add(rng)
+                        db_api.mac_range_update_set_full(context, rng)
                         LOG.info("MAC range {0} is full".format(rng["cidr"]))
                         continue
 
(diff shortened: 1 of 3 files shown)


quark/ipam.py
```diff
@@ -222,12 +221,12 @@ def allocate_mac_address(self, context, net_id, port_id, reuse_after,
                         next_address = mac_address
                     else:
                         next_address = rng["next_auto_assign_mac"]
-                        next_auto = next_address + 1
-                        if next_auto > last:
-                            next_auto = -1
-                        db_api.mac_address_range_update(
-                            context, rng, next_auto_assign_mac=next_auto)
-
+                        if next_address + 1 >= rng["last_address"]:
```
asadoughi: Switched from `>` to `>=`?

Cerberus98: Yeah, it's a bit confusing. See how I also removed the +1 on line 226?

Now we're using the value as it exists in the table when we read it, and THEN updating next_auto_assign_mac.

asadoughi: Previously it read `next_auto > last` where `next_auto = next_address + 1` which is the same as `next_address + 1 > last`. Now, it reads `next_address + 1 >= rng["last_address"]` which is `next_address + 1 > last`. So, `next_address + 1 > last` is different from `next_address + 1 >= last`.

Cerberus98: Ahh, I see. Yeah, ok.

Cerberus98:
Updated, and added a test to allocate the next to last address. Test failed before I changed >= to >

asadoughi:
LGTM :100:

PR #351, opened 2015-03-09
Validates security group name and description

Cerberus98:
RM11507

Security group names and descriptions are now correctly limited to 255
characters a piece, and an error will be thrown to the user if either is
greater.

### DIFF
--- a/quark/plugin_modules/security_groups.py
+++ b/quark/plugin_modules/security_groups.py
@@ -27,6 +27,8 @@ from quark import protocols
 CONF = cfg.CONF
 LOG = logging.getLogger(__name__)
 DEFAULT_SG_UUID = "00000000-0000-0000-0000-000000000000"
+GROUP_NAME_MAX_LENGTH = 255
+GROUP_DESCRIPTION_MAX_LENGTH = 255
 
 
 def _validate_security_group_rule(context, rule):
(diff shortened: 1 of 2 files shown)


quark/plugin_modules/security_groups.py
```diff
@@ -63,13 +65,28 @@ def _validate_security_group_rule(context, rule):
...
+    if "name" in security_group:
+        if len(security_group["name"]) > GROUP_NAME_MAX_LENGTH:
+            raise exceptions.InvalidInput(msg="Group name must be 255 "
+                                              "characters or less")
+
+        if security_group["name"] == "default":
+            raise sg_ext.SecurityGroupDefaultAlreadyExists()
+
+    if ("description" in security_group and
+            len(security_group["description"]) > GROUP_DESCRIPTION_MAX_LENGTH):
```
jmeridth: why use an `and` on this one and not the one before?  *nit*

jmeridth:
:+1: one little nit, will merge if you want :smile:

quark/plugin_modules/security_groups.py
Cerberus98: Because there are two conditions I needed to check under "name" being in security group. Does that make sense?

insequent:
LGTM :snowman:

PR #352, opened 2015-03-10
Fixed gunicorn run script

roaet:
RM6846


insequent:
Reviewed roaet's PR :+1:

PR #354, opened 2015-03-13
Adds quotas for fixed IPs on ports

Cerberus98:
RM11643

Implements quota checking the number of fixed IP addresses allowed on a
port on create and update port, as well as create and update IP Address.


Cerberus98:
Awesome, a new day, another breaking namespace change upstream that doesn't solve any real problems.

Cerberus98:
Once Justin's fix passes tests finally and merges I'll rebase this on that

mpath:
:+1:

PR #357, opened 2015-03-16
Enable container-based travis-ci builds

asadoughi:
Seeing if http://docs.travis-ci.com/user/workers/container-based-infrastructure/ let's us beat the wait for non-container builds.


insequent:
Sounds fun. LGTM :+1:

PR #358, opened 2015-03-16
Cleaning up _model_query and adding SG attribs

insequent:
RM11464

Had initial ambitions of making this dynamic (hence the branch name), but decided against it in the end.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
   - "2.7"
-  - "2.6"
 before_install:
   - "export NO_EVENTLET=1"
 install:
--- a/quark/tests/test_db_api.py
+++ b/quark/tests/test_db_api.py
@@ -15,10 +15,13 @@
 
 import mock
 import netaddr
+from oslo_log import log as logging
 
 from quark.db import api as db_api
 from quark.db import models
 from quark.tests.functional.base import BaseFunctionalTest
+
+LOG = logging.getLogger(__name__)
 
 
 class TestDBAPI(BaseFunctionalTest):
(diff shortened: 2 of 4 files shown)


Cerberus98:
Seems legit

jmeridth:
tests?

insequent:
Good point, I'll add some tests.

jmeridth:
Nice refactoring btw.  Forgot to say that in my last comment :smile:

insequent:
Ok, it took a bit to work out the quirks, but I made the filter generation dynamic. This also fixes the bug where if you supply an attribute that exists on a different model, but not the one you're filtering for, Quark raises and Neutron returns a 500.

NOTE: I also removed py26 testing.

jmeridth:
Having a little Ruby metaprogramming flashback, but it is concise.  :+1:   Nice job adding tests.

quark/tests/test_db_api.py
```diff
@@ -130,6 +133,67 @@ def test_ip_address_find_ip_address_list(self):
         except Exception as e:
             self.fail("Expected no exceptions: %s" % e)
 
+    def test_model_query(self):
```
jmeridth: I'd prefer to have each of these groupings in their own test method.  It makes the test count bigger....jk.  Just seems cleaner to me.

insequent:
ORMs are a bit meta. So, when I first tried to make this dynamic, it felt like I was creating a secondary mapping outside of models, i.e. a second source of truth. Hence why I backed off. However in the end, this feels like the cleanest way to go. The only other thing I could think to try was to put this function inside models, to at least keep everything in one place, but I'm don't know how much that would help.

Thanks for the suggestion on separate tests. I was thinking the same thing last night and totally forgot.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: python
 python:
   - "2.7"
-  - "2.6"
```
asadoughi: If the dict comprehension is not going to work in travis, it's not going to work in `tox.ini` either. Alternatively, you can use the older style: `d = dict((key, value) for (key, value) in iterable)` to stay 2.6 compatible.

asadoughi: Oops, didn't see tox.ini at the bottom there.

asadoughi:
Where are we on the 2.6 decision? I thought we were still on the fence about removing it. I don't recall any conclusions from yesterday's IRC conversation.

Cerberus98:
Much cleaner :heart:

jmeridth:
Removal is fine as upstream doesn't support it any longer re: py26

PR #359, opened 2015-03-17
Wraps delete_vif in exception handling

Cerberus98:
In the event that Redis has gone missing/down, deleting a port would
fail as a result of trying to clean up what may or may not be in Redis.
Continuing in the theme of "It's cheaper to talk to redis than the db"
we simply log any exception from failing to clean up redis and get on
with deleting the port. Meanwhile, the only side-effect of any data
continuing to live in redis is meaningless cruft, and the redis_sg_tool
can find and purge orphaned data later if it's deemed a problem.


roaet:
Sweet. Yay feature quarantines! :hatching_chick:

PR #361, opened 2015-03-18
Cache allocation pools

mpath:
Adds logic for a cache of allocation pools in the Subnet model, _allocation_pool_cache
column. Stored as a json string and invalidated on update_subnet or update_subnets.

RM#10652


quark/plugin_modules/subnets.py
```diff
@@ -268,11 +273,12 @@ def update_subnet(context, id, subnet):
             alloc_pools = allocation_pool.AllocationPools(subnet_db["cidr"],
                                                           allocation_pools)
 
+        LOG.info("Performing quota limit check on alloc pools/subnet.")
         quota.QUOTAS.limit_check(
             context,
             context.tenant_id,
             alloc_pools_per_subnet=len(alloc_pools))
-
+        LOG.info("Check passed.")
```
insequent: With transaction IDs in the logs, this isn't a huge issue, but I think it'll probably be good  to include the check name that's passing here (like `alloc pools/subnet quota check passed`.

quark/plugin_modules/subnets.py
```diff
@@ -344,8 +358,12 @@ def get_subnet(context, id, fields=None):
     net_id = subnet["network_id"]
     net_id = STRATEGY.get_parent_network(net_id)
     subnet["network_id"] = net_id
-
-    return v._make_subnet_dict(subnet)
+    retval = v._make_subnet_dict(subnet)
+    cache = retval.get("_allocation_pool_cache")
+    if not cache:
+        new_cache = v._allocation_pools(subnet)
```
asadoughi: This seems odd, too. Since `_allocation_pools` leads with an underscore it is a private function by notation and shouldn't called outside of its module. It looks like `allocation_pools()` would make more sense as a method of `quark.db.models.Subnet` where it would load the cached response if it was valid. Thoughts?

quark/plugin_modules/subnets.py
```diff
@@ -344,8 +358,12 @@ def get_subnet(context, id, fields=None):
     net_id = subnet["network_id"]
     net_id = STRATEGY.get_parent_network(net_id)
     subnet["network_id"] = net_id
-
-    return v._make_subnet_dict(subnet)
+    retval = v._make_subnet_dict(subnet)
+    cache = retval.get("_allocation_pool_cache")
```
asadoughi: This seems odd. Why are we checking this on the view's `_make_subnet_dict` output and not the subnet object?

quark/plugin_modules/subnets.py
asadoughi: If you are going to add log statements, I'd prefer if you were more verbose. In a multiprocess/thread context, it would make it a lot easier to parse if we could discriminate between different log statements, i.e. I wouldn't be able to easily tell the difference between this line and line 309. Suggestion: "Check for quota limit on alloc pools/subnet passed".

quark/plugin_modules/subnets.py
```diff
@@ -83,6 +83,11 @@ def _validate_subnet_cidr(context, network_id, new_subnet_cidr):
             raise exceptions.InvalidInput(error_message=err_msg)
 
 
+def set_allocation_pool_cache(context, subnet, cache_data=None):
+    LOG.debug("Updating allocation pool cache to %s" % cache_data)
+    db_api.subnet_update_set_alloc_pool_cache(context, subnet, cache_data)
```
asadoughi: Nit: This function looks useless.

quark/plugin_modules/subnets.py
asadoughi: Oops, you beat me to it since I was still drafting my comments.

asadoughi:
Finished reviewing first draft. One nit and a concern with where the `allocation_pools` function is placed.

quark/plugin_modules/subnets.py
```diff
@@ -318,6 +321,9 @@ def update_subnet(context, id, subnet):
                 ip_policies.ensure_default_policy(cidrs, [subnet_db])
                 subnet_db["ip_policy"] = db_api.ip_policy_update(
                     context, subnet_db["ip_policy"], exclude=cidrs)
+                # invalidate the cache
+                LOG.info("Invalidating allocation pool cache.")
```
asadoughi: Nit: two log statements in a row with the log statement inside `set_allocation_pool_cache`. Eliminate one, please.

asadoughi:
I've pushed my suggested changes https://github.com/rackerlabs/quark/compare/cache_allocation_pools...asadoughi:cache_allocation_pools it works except for 3 UT. I'm not sure what the last 3 are hung up on, but I'm sure you can figure it out.

asadoughi:
3 UT were fixed in my suggested branch.

asadoughi:
:+1: LGTM. As co-author I don't feel that I should merge this.

insequent:
Into the model and out of the view. LGTM :+1:

PR #362, opened 2015-03-18
C-A-S IPs

asadoughi:
RM11014


quark/db/api.py
```diff
@@ -313,10 +314,73 @@ def ip_address_find(context, lock_mode=False, **filters):
...
+    return transaction_id if row_count > 0 else None
+
+
+def ip_address_reallocate_find(context, transaction_id):
+    address = ip_address_find(context, transaction_id=transaction_id,
+                              scope=ONE)
+    if not address:
+        LOG.warn("Couldn't find IP address with transaction_id %s",
+                 transaction_id)
+        return
```
Cerberus98: This seems fine, though I'm trying to think of a situation where we got this far and suddenly can't find the IP. Better safe than sorry, of course.

quark/db/sqlalchemy_adapter.py
```diff
@@ -0,0 +1,23 @@
+from sqlalchemy.orm.persistence import BulkUpdate
+from sqlalchemy import sql
+
+
+class BulkUpdateMySqlLimit(BulkUpdate):
+    def __init__(self, query, values, **kwargs):
+        super(BulkUpdateMySqlLimit, self).__init__(query, values)
+        self.kwargs = kwargs
+
+    def _do_exec(self):
```
Cerberus98: Love those magic sqlalchemy methods

asadoughi:
Added transactions table, still needs lots of tests added/fixed.

asadoughi:
I've got unit tests to pass, but I'm still fixing them a bit.  Also, realized I forgot about `address_type` in my original pass, so I'll need to add a functional test for that. So, please do not merge yet. Thanks.

asadoughi:
Two tests still failing on travis-ci that don't fail locally for me.

asadoughi:
Hmm, 2 tests don't fail with tox, but definitely fail when running the `nosetests` travis.yml command.

asadoughi:
Oh, I think it's from setting up the database engine. 

This passes tests:
```
nosetests --with-coverage --cover-package=quark --cover-erase --cover-html --cover-html-dir=.cover-report --cover-min-percentage=90 -- quark.tests.functional.mysql
```

Whereas this doesn't:
```
nosetests --with-coverage --cover-package=quark --cover-erase --cover-html --cover-html-dir=.cover-report --cover-min-percentage=90
```

asadoughi:
Yeah, it looks like mixing MySQL and SQLite backends in one process with global singletons spread throughout is not going to work.

asadoughi:
OK. Got tests to pass by moving `.travis.yml` to run `tox` as we do locally.

quark/ipam.py
```diff
@@ -352,104 +352,68 @@ def attempt_to_reallocate_ip(self, context, net_id, port_id, reuse_after,
...
-            get_policy = models.IPPolicy.get_ip_policy_cidrs
-
             try:
                 with context.session.begin():
-                    # NOTE(mdietz): Before I removed the lazy=joined, this
-                    #               raised with an unknown column "address"
-                    #               error.
-                    address = db_api.ip_address_find(elevated, **ip_kwargs)
+                    transaction = db_api.transaction_create(context)
+                m = models.IPAddress
```
jmeridth: this line and below doesn't need to be under the `context.session.begin()`?  all we need that for is the transaction itself and use the `id` in the reallocated ip address?  Just trying to understand. :smile:

jmeridth: "line and below" == line 389 and below.  does it need to be nested under `context.session.begin()`

jmeridth: talking with @Cerberus98, does the new `sqlalchemy_adapter` usage take care of this? https://github.com/rackerlabs/quark/pull/362/files#diff-405aeec781a4833096de273adf9a2cd8R332

asadoughi: Good question. The way I understood the algorithm design is that we insert a Transaction row, flush it (happens on the `with` block's `__exit__`) and then use that Transaction row's `id` to atomically reallocate a specific deallocated IP via `UPDATE..WHERE..LIMIT 1`. So, the newly inserted Transaction row doesn't need to be in a `context.session.begin`, I could have used `context.session.flush()`. For example,

```
transaction = db_api.transaction_create(context)
self.context.session.flush()
```

In other words, since `UPDATE..WHERE..LIMIT 1` is an atomic operation, we don't need to use the database's transaction of `BEGIN` .. `END`.

@Cerberus98 can correct me if I'm wrong. :)

quark/db/api.py
```diff
@@ -314,10 +316,72 @@ def ip_address_find(context, lock_mode=False, **filters):
...
 
 
 @scoped
+def ip_address_reallocate(context, update_kwargs, **filters):
+    LOG.debug("ip_address_reallocate %s", filters)
+    query = context.session.query(models.IPAddress)
+    model_filters = _model_query(context, models.IPAddress, filters)
+    query = query.filter(*model_filters)
+    row_count = quark_sa.update(query, update_kwargs,
+                                update_args={"mysql_limit": 1})
```
asadoughi: Yes, this is where the `UPDATE..LIMIT 1` operation happens. It will be available in sqlalchemy==1.0.0, so until we adopt that release or newer we'll have to use the adapter.

jmeridth:
:+1:

asadoughi:
I'm having some difficulty running `neutron-server` locally. It's not calling into the plugin on the few requests I've did.

asadoughi:
Nevermind about neutron-server. I got it working -- I forgot some key headers in my curl's.

asadoughi:
Other than reverting some of #368, my integration testing went well.

quark/db/api.py
```diff
@@ -314,10 +316,72 @@ def ip_address_find(context, lock_mode=False, **filters):
...
+
+    LOG.info("Potentially reallocatable IP found: "
+             "{0}".format(address["address_readable"]))
+    subnet = address.get('subnet')
+    if not subnet:
+        LOG.debug("No subnet associated with address")
+        return
+    if subnet["do_not_use"]:
+        LOG.debug("Subnet marked as do_not_use")
+        return
```
Cerberus98: This implies we need a way to an IP as unusable sooner rather than later. Seems like in this scenario we'd run into these IPs (there'd probably be more than one) and never get past them.

quark/db/api.py
```diff
@@ -314,10 +316,72 @@ def ip_address_find(context, lock_mode=False, **filters):
...
+
+    # TODO(amir): performance test replacing this with SQL in
+    #             ip_address_reallocate's UPDATE statement
+    cidr = netaddr.IPNetwork(address["subnet"]["cidr"])
+    if addr not in cidr:
+        LOG.info("Address {0} isn't in the subnet "
+                 "it claims to be in".format(
+                     address["address_readable"]))
+        context.session.delete(address)
```
Cerberus98: As above, since we're carrying around this madness into new code paths. 

Not saying we need it in this patch, just realizing we still haven't added it.

quark/db/api.py
```diff
@@ -760,3 +824,9 @@ def ip_policy_update(context, ip_policy, **ip_policy_dict):
 
 def ip_policy_delete(context, ip_policy):
     context.session.delete(ip_policy)
+
+
+def transaction_create(context):
+    transaction = models.Transaction()
+    context.session.add(transaction)
+    return transaction
```
Cerberus98: Woo!

quark/db/migration/alembic/versions/5632aa202d89_added_transactions_table.py
```diff
@@ -0,0 +1,36 @@
...
+                    sa.Column('id', sa.Integer(), nullable=False),
+                    sa.PrimaryKeyConstraint('id'),
+                    mysql_engine='InnoDB')
+    op.add_column(u'quark_ip_addresses',
+                  sa.Column('transaction_id', sa.Integer(), nullable=True))
+    op.create_foreign_key('fk_quark_ips_transaction_id',
+                          'quark_ip_addresses',
+                          'quark_transactions',
+                          ['transaction_id'],
+                          ['id'])
```
Cerberus98: I'd like us to add a status or message field here at some point (again, not this patch). I think it would be useful for actually following what went on. We could also theoretically use it to move forward with async allocations. In other words, some external processes could looking for all started transactions that aren't completed, and try to finish them.

Just thinking out loud.

quark/ipam.py
Cerberus98: I kind of get the impression that the expression language adapter you have above is implicitly emitting the begin and end on the reallocate call (I haven't run this code yet, but I'm assuming you showed it works :-) )

quark/db/api.py
asadoughi: Is that a necessarily a new feature for upcoming IP address efforts? Or, do you mean to eschew a join on subnets to filter out those with `do_not_use = True`.

Cerberus98:
Looks squared away. Does it work? :-)

quark/db/api.py
Cerberus98: I actually didn't think about this the right way. From the looks of it, it would mark the IP with the transaction ID, and never remove it. Am I reading that right?

asadoughi: Oh, I remember now, are you referring to the `skipped` column we talked about months ago?

quark/db/api.py
asadoughi: As noted in my TODO comments above, we could expand our UPDATE query instead of an additional query. Alternatively, I'd be open to the `skipped` column idea we talked about ages ago.

quark/db/migration/alembic/versions/5632aa202d89_added_transactions_table.py
asadoughi: Makes sense. Adding it when we have a real use for it is the way to go.

quark/db/api.py
Cerberus98: Yep, that guy

asadoughi:
Beyond the unit and functional tests, I did a few POST /ports requests and saw the neutron-server log messages about IP reallocation go through. I was going to rely on our fantastic perf QE team to give us the real benchmarks. I could try giving it some load locally to see if it falls over. I'll comment this PR shortly thereafter.

quark/db/api.py
asadoughi: > I actually didn't think about this the right way. From the looks of it, it would mark the IP with the transaction ID, and never remove it. Am I reading that right?
Yes, why would you need to unset the transaction ID? One concern I can think of is the quark_transactions table having too many rows.

asadoughi:
In my very unscientific test, it was 2x faster (18 serial POST /ports requests on production data). The next bottleneck appears to be the `SELECT..FOR UPDATE` on  `select_subnet` when allocating a v6 address.

PR #363, opened 2015-03-19
Removing XML serialization code

insequent:
The class no longer exists upstream. RM11046

### DIFF
--- a/quark/tests/api/extensions/test_ports_quark.py
+++ b/quark/tests/api/extensions/test_ports_quark.py
@@ -1,0 +1,44 @@
+# Copyright 2015 Openstack Foundation
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+#  under the License.
+
+import json
+import mock
+from oslo_log import log as logging
+
+from quark.api.extensions import ports_quark
+from quark.tests.functional.base import BaseFunctionalTest
+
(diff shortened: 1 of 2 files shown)


jmeridth:
:+1:

quark/tests/api/extensions/test_ports_quark.py
```diff
@@ -0,0 +1,48 @@
...
+
+        def mock_fun(context, id, body):
+            result = {"id": id, "body": body}
+            return result
+
+        handler = ports_quark.QuarkPortsUpdateHandler(mock_plugin)
+        self.assertEqual(handler._plugin, mock_plugin)
+
+        mock_plugin.post_update_port.side_effect = mock_fun
+        mock_request.context = mock.MagicMock()
```
asadoughi: This isn't necessary. MagicMock will return MagicMock() all it's non-class-specific attributes all the way down.

quark/tests/api/extensions/test_ports_quark.py
```diff
@@ -0,0 +1,48 @@
...
+        port_id = "I_am_a_port"
+
+        def mock_fun(context, id, body):
+            result = {"id": id, "body": body}
+            return result
+
+        handler = ports_quark.QuarkPortsUpdateHandler(mock_plugin)
+        self.assertEqual(handler._plugin, mock_plugin)
+
+        mock_plugin.post_update_port.side_effect = mock_fun
```
asadoughi: Consider doing `mock_plugin.post_update_port.return_value = {"id": id, "body": body}` instead of an extra function.

quark/tests/api/extensions/test_ports_quark.py
insequent: Yeah, I was just being explicit.

quark/tests/api/extensions/test_ports_quark.py
insequent: That would be cleaner. I originally thought there was a lot more going on in post_update_port.

insequent:
There, nice and slim @asadoughi.

asadoughi:
LGTM. The whole serialization thing when it was first implemented was kinda hackish method I copied from an existing upstream implementation. i.e. I don't think extension code should be worried about serializers at all. Perhaps, since then there's been a better way to implement what the extension is trying to do.

PR #364, opened 2015-03-19
Adding oslo.log requirement

insequent:
This is causing jenkins tests to fail. RM11046


asadoughi:
Shouldn't it be pulled in by neutron?

insequent:
Theoretically. Maybe it's failing for a different reason. I'll check again.

insequent:
Oh versioning, confused myself on this one. We have an old version of neutron on the tests.

PR #365, opened 2015-03-20
Fixup for SG names being too long exceptions

Cerberus98:
RM11507

Updates the use of InvalidInput when validating security group names and
descriptions to use the "error_message" keyword rather than "msg."


mpath:
:+1:

PR #366, opened 2015-03-23
Adds validation around port create with fixed ips

Cerberus98:
RM11016

Updates port creation to add the following validations that were
missing:

* Creating a port on a providernet as a non-admin now gives a proper
  not authorized message
* Creating a port with a subnet_id that isn't owned by the network
* Creating a port with subnet_id that doesn't exist

This also fixes a case where a tenant trying to use another tenant's
subnet in the request receives an unhelpful "Out of IPs" message. The
code now correctly returns Not Found.


asadoughi:
LGTM

PR #367, opened 2015-03-23
Initial work on adding easier to parse logs

roaet:
RM11409

### DIFF
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -20,6 +20,7 @@ Quark Pluggable IPAM
 import functools
 import itertools
 import random
+import time
 import uuid
 
 import netaddr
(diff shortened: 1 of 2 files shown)


quark/ipam.py
```diff
@@ -119,7 +120,71 @@ def generate_v6(mac, port_id, cidr):
...
+class QuarkIPAMLog(object):
+    def __init__(self):
+        self.entries = {}
+        self.success = True
+
+    def make_entry(self, fx_name):
+        if fx_name not in self.entries:
+            self.entries['fx_name'] = []
+        entry = QuarkIPAMLogEntry(self, fx_name)
+        self.entries['fx_name'].append(entry)
```
asadoughi: Did you mean to quote `fx_name`? Should it be `self.entries[fx_name]`?

Sidenote: `self.entries = collections.defaultdict(list)` would save a line.

quark/ipam.py
```diff
@@ -119,7 +120,71 @@ def generate_v6(mac, port_id, cidr):
...
+
+    def get_time(self):
+        if not hasattr(self, 'end_time'):
+            return 0
+        return self.end_time - self.start_time
+
+
 class QuarkIpam(object):
+    def __init__(self):
+        self.ipam_log = QuarkIPAMLog()
```
asadoughi: Is `QuarkIpam` instantiated as a global singleton? If so, will `make_entry` need to be thread safe?

asadoughi:
Finished reviewing first draft. Cool stuff. Unit tests, please.

roaet:
Unit tests of log statements?

quark/ipam.py
roaet: Indeed.

asadoughi:
Unit tests of `QuarkIPAMLog` and `QuarkIPAMLogEntry`, yes, please. For example, something that would have caught the `fx_name` quoting.

roaet:
I will fix the fx_name quoting. And still, probably not going to write tests for it. While you are definitely right on that error being caught, I'm not going to write the tests for log statements.

quark/ipam.py
roaet: I will need to import it (thus adding a line). For "one line" that I will lose by importing? Seems pretty silly and unperformant.

quark/ipam.py
roaet: Checking on this. Not familiar with ipam enough to know if this is a thing or if it even matters.

roaet: QuarkIPAM is not a singleton (from dietz). I also don't think this being 'thread-safe' matters.

quark/ipam.py
asadoughi: I just thought I would mention it if you hadn't heard of the `defaultdict` before :). P.S. importing +1 line, removing if statement for contains -1 line, removing assignment of list -1 line = 1 - 1 - 1 = -1 line in total.

asadoughi:
OK, that's fine, regarding unit tests if you feel that way. In the future, if another dev wanted to end up expanding on this awesome functionality you're adding, they'd probably end up writing regression tests for it since we wouldn't want to break it once it was being depended on and working.

quark/ipam.py
Cerberus98: I can't think of a reason it actually needs to be a singleton, for what it's worth, especially since I don't deliberately remember doing it that way.

Cerberus98: To put it another way, it's clear we need to change this, but I think the best approach is to create the actual logging object where we intend to use it and pass it around, or shove it in the context variable since that's kind of what it's for anyway. Making it thread safe via some other means still won't stop competing calls logging from bleeding together.

asadoughi: Sounds like a good approach to me.

jmeridth:
As this isn't just "log statements", it is an object that manages them, I have to agree with @asadoughi  on this one.  There are some easy, regressable comfort, unit testing opportunities here. :smile: Please write some tests that cover any non-basic logic.

quark/ipam.py
```diff
@@ -119,6 +120,72 @@ def generate_v6(mac, port_id, cidr):
...
 
+def ipam_logged(fx):
+    def wrap(self, *args, **kwargs):
+        log = QuarkIPAMLog()
+        kwargs['ipam_log'] = log
+        try:
+            return fx(self, *args, **kwargs)
+        finally:
+            log.end()
+    return wrap
```
asadoughi: For fun: Another python pattern to accomplish the same thing I've recently gleaned: http://stackoverflow.com/a/1984346/2529333

roaet: Thanks for that alternative. maybe next time

PR #368, opened 2015-03-26
Removing POST update logic

insequent:
Also sending PATCH requests through update_port, as there's no definitive
difference between PATCH and PUT updates yet. RM11046

I'm completely open to suggestions on this one. Just trying to clean things up some.


quark/api/extensions/ports_quark.py
```diff
@@ -15,7 +15,8 @@
 
 from neutron.api import extensions
 from neutron import manager
-from neutron import wsgi
+from neutron.wsgi import JSONDeserializer
+from neutron.wsgi import JSONDictSerializer
```
jmeridth: why not

`from neutron.wsgi import JSONDeserializer, JSONDictSerializer`

it's early, so forgive if I'm brain farting

insequent: Flake8 yells at me :/

insequent: I really wanted to get out of the serialization game entirely here, but I couldn't find a better way to do this. I think ultimately, this kind of change should reside in Neutron itself.

jmeridth: ah, flake8, y u so silly?  :stuck_out_tongue_winking_eye:  interesting, I wonder why.  will try locally.

upstream patch submission.......ehhhhh? :wink:

still reviewing the rest.

jmeridth:
TIL about HTTP PATCH: http://tools.ietf.org/html/rfc5789

:+1:   LGTM

quark/plugin_modules/ports.py
```diff
@@ -382,58 +382,6 @@ def update_port(context, id, port):
...
-        address["deallocated"] = 0
-
-        already_contained = False
-        for port_address in port_db["ip_addresses"]:
-            if address["id"] == port_address["id"]:
-                already_contained = True
-                break
-
-        if not already_contained:
-            port_db["ip_addresses"].append(address)
```
asadoughi: Also, `update_port` is replacing all `fixed_ips` with one specified rather than appending like this here.

quark/plugin_modules/ports.py
```diff
@@ -382,58 +382,6 @@ def update_port(context, id, port):
...
-
-    port = port["port"]
-    if "fixed_ips" in port and port["fixed_ips"]:
-        for ip in port["fixed_ips"]:
-            address = None
-            ipam_driver = ipam.IPAM_REGISTRY.get_strategy(
-                port_db["network"]["ipam_strategy"])
-            if ip:
-                if "ip_id" in ip:
-                    ip_id = ip["ip_id"]
```
asadoughi: Has the requirement to associate an IP address to a port via an IP address resource's uuid been removed? I don't think `update_port` handles this case.

jmeridth:
"If the Request-URI does not point to an existing resource, the server MAY create a new resource, depending on the patch document type (whether it can logically modify a null resource) and permissions, etc."

It's like a built-in ORM post or update call, niiiiiiice.

asadoughi:
I don't think you can replace `post_update_port` with `update_port` since one is doing the POST equivalent, replacing all sub-resources/attributes, and the other is doing a PATCH equivalent, appending to the existing sub-resources/attributes.

jmeridth:
@asadoughi doesn't the PATH method create the sub-resources/attributes if they don't exist?  Be aware, I'm still reading up on it.

asadoughi:
To phrase it another way, this change doesn't look like it's fulfilling the requirements the previous implementation had.

asadoughi:
(We may have been abusing HTTP in the previous implementation to meet business requirements, but there were specific business requirements).

insequent:
This is exactly the kind of feedback I was looking for. Let me work on your 2 concerns @asadoughi. I still need to incorporate some fixes for update_port as well, but I was separating it out for clarity.

insequent:
@asadoughi: That does make me wonder though, if we really want PATCH to simply append. My instincts tell me that we should allow empty updates to clear a field.

quark/plugin_modules/ports.py
insequent: First thoughts, why is this not in update_port?

asadoughi: No specific reason. I sent you some internal docs via IRC message that may help.

insequent:
It was decided to remove the POST update logic altogether, so dropping the PATCH code.

insequent:
To be fair, it's probably smart to wait for Neutron to add in this functionality (less room for error on our part, as extensions do JSON parsing, etc.). Cleaner this way as well.

quark/api/extensions/ports_quark.py
```diff
@@ -1,99 +0,0 @@
...
-
-RESOURCE_NAME = "port"
-RESOURCE_COLLECTION = RESOURCE_NAME + "s"
-EXTENDED_ATTRIBUTES_2_0 = {
-    RESOURCE_COLLECTION: {
-        "network_id": {"allow_post": True, "default": '',
-                       "is_visible": True},
-        "tenant_id": {"allow_post": True, "default": '',
-                      "is_visible": True},
-        "segment_id": {"allow_post": True, "default": False},
```
asadoughi: Whoops, we need this.

PR #370, opened 2015-04-08
RM12111 Made jarvis run tests like tox

asadoughi:
Replaced run_tests.sh with commands that run with tox. I would use
tox itself here, but it wouldn't install the correct branch of
neutron.


run_tests.sh
```diff
@@ -1,254 +1,8 @@
...
-# not when we're running tests individually. To handle this, we need to
-# distinguish between options (testopts), which begin with a '-', and
-# arguments (testargs).
-if [ -z "$testargs" ]; then
-  if [ $no_pep8 -eq 0 ]; then
-    run_pep8
-  fi
-fi
+# TODO: make tox install the correct version of neutron so we can run tox
+#       instead of these commands manually
```
asadoughi: We should probably move our requirements.txt to look something like this so I don't have to work around tox: https://github.com/rackerlabs/ironic-neutron-plugin/blob/master/requirements.txt#L8 In other words, version our upstream dependencies on a specific commit.

Cerberus98:
Looks good to me.

PR #372, opened 2015-04-15
Compare-and-swap: MAC address reallocations

asadoughi:
* Removed ORDER BY clause since SQLAlchemy doesn't support ORDER BY
on UPDATEs.
* Added migration for addding transaction_id column to quark_mac_addresses.

RM11013


asadoughi:
travis-ci doesn't seem to be recognizing my force push. Closing this PR and opening new one with same commit.

PR #375, opened 2015-04-20
Shuffle list of sentinels to distribute connection load

asadoughi:
RM12113

### DIFF
--- a/quark/cache/redis_base.py
+++ b/quark/cache/redis_base.py
@@ -15,6 +15,7 @@
 
 import functools
 import json
+from random import shuffle
 import string
 
 import netaddr
@@ -136,6 +137,9 @@ class ClientBase(object):
     def _compile_sentinel_list(self):
         self._sentinel_list = [tuple(host.split(':'))
                                for host in CONF.QUARK.redis_sentinel_hosts]
+        # NOTE(asadoughi): RM12113: shuffle list of sentinels to distribute
+        #                           connection load
+        shuffle(self._sentinel_list)
         if not self._sentinel_list:
             raise TypeError("sentinel_list is not a properly formatted"
                             "list of 'host:port' pairs")


quark/cache/redis_base.py
```diff
@@ -136,6 +137,9 @@ def _ensure_connection_pools_exist(self):
     def _compile_sentinel_list(self):
         self._sentinel_list = [tuple(host.split(':'))
                                for host in CONF.QUARK.redis_sentinel_hosts]
+        # NOTE(asadoughi): RM12113: shuffle list of sentinels to distribute
+        #                           connection load
+        shuffle(self._sentinel_list)
```
tr3buchet: this does nothing

asadoughi: Please elaborate. `shuffle` is an in-place operator.

```
>>> x = range(5)
>>> shuffle(x)
>>> x
[3, 0, 1, 2, 4]
```

tr3buchet: yes, I know what shuffle does. The order of the sentinel list is unimportant.

asadoughi: Ok, perhaps I've misread the redis-py library, how is the sentinel the agent connects to determined? It's not the first element of this list?

tr3buchet: the client connects to them all, and it even has a configurable minimum that must be able to connect to else fail. If what you need is for the actual read slave to be randomized, the SentinelConnectionPool does this for you.

asadoughi: (1) In practice, I only see the agent connected to one sentinel via netstat. Admittedly, this is much after the agent has gotten into it's rhythm and supposedly asked for all the masters and slaves it would ever need.
(2) The minimum number of sentinels configuration parameter refers to the number of peers the sentinel that's being queried sees. https://github.com/andymccurdy/redis-py/blob/master/redis/sentinel.py#L199
(3) I see the read slaves being randomized, but I don't see the list of sentinels being randomized.  https://github.com/andymccurdy/redis-py/blob/master/redis/sentinel.py#L179 versus https://github.com/andymccurdy/redis-py/blob/master/redis/sentinel.py#L109

asadoughi: I imagine the only way agents are connecting to the second (or beyond) element of the sentinel list is when they hit ConnectionError: https://github.com/andymccurdy/redis-py/blob/master/redis/sentinel.py#L214-L220 and the list is then modified to move the succesfully-connected-to-sentinel as the first in the list.

tr3buchet: the sentinels are rotated through as shown [here](https://github.com/andymccurdy/redis-py/blob/master/redis/sentinel.py#L217-L221) and in `discover_slaves` below it. sounds like you are keeping the connection open permanently in the quark agent? Might make more sense to choose a new slave/sentinel every few minutes or however often it runs to distribute load.

asadoughi: My reading of `discover_master` is as follows: No rotating in the sentinel list occurs if the conditions are met successfully by the first sentinel in the list. If the first sentinel in the list does _not_ meet the conditions, then the nth sentinel in the list is swapped with the first element when the nth sentinel meets the conditions. (Followed by: Return the state from the nth sentinel. Otherwise if no master is found, raise an error.)

The redis-py library is keeping a connection alive to the redis-sentinel via connection pool in the quark-agent because `redis.sentinel.Sentinel` never explicitly disconnects from its current sentinel, `self.sentinels[0]`. I would have to subclass `redis.sentinel.Sentinel` and override `discover_master` and `discover_slaves` to get the desired outcome of actually rotating through sentinels. I prefer the solution in this PR since it's much lower complexity, making it easier to maintain and comprehend.

tr3buchet: I wonder why it rotates the sentinels when it discovers the master node but not when discovering the slaves. I'm putting together a client implementation and for the read only case that seems like it would be useful. Also the connection handling for long running processes is important so I'm trying to figure that best way to manage that. 

Is this code used by quark agent or just neutron/quark?

asadoughi:
The sentinel list is not modified in `discover_master` the majority of the time anyway. The more I read redis-py's source code the less I like it. So, for the short term, the shuffling operation I've added in this PR will mitigate the production issue by changing which sentinel is first in the list at runtime.

I'm open to discussing other client implementations outside of this PR. As I've already stated, I don't think building on top of redis-py is a sound engineering decision. If you want to build something alternative to redis-py, that's fine, but I don't think this PR should be blocked on that.

tr3buchet:
It only rotates the sentinels when one fails to provide a good master. In the case of discover master, it's putting the sentinel with good master state at the front of the list. It doesn't load balance those suckers, it just keeps using a known good one each time until that changes.

So 2 questions:
1. is this code used by neutron/quark or the quark agent? 
2. how long do you see the sentinel connection being maintained?

asadoughi:
OK, glad we're on the same page for `discover_master`.

1. Both neutron-server and quark-agent use this base class. Both neutron-server and quark-agent use both masters and slaves for writing and reading respectively.
2. quark-agent maintains one connection forever to one redis-sentinel. I haven't checked neutron-server.

tr3buchet:
alright checkit! So I've been playing around with the client/sentinel and there is no reason to keep that sentinel connection alive beyond the first time you use it to discover master or slaves. disconnection from it is a bit arcane, but it works fine. In future, if the connection to the master or slave goes down and you need to reconnect, the sentinel connection is reestablished without issue. If all of our sentinels are good and in the same order on all clients, we're using the 1 sentinel to manage all the connections. That is actually fine, because the client will failover if it goes down, and the sentinel can manage that many connections, but if we close the sentinel session, the clients will only be connected for milliseconds at a time, easing connection strain across the board. Long story short, might as well shuffle them here for now.

asadoughi:
Yes, that sounds like a better client/sentinel design than how redis-py works today. I don't see much velocity in the redis-py repo's pull requests so I'm skeptical about making improvements to it. For example, I haven't heard back from a PR I filed to fix the bug that causes our neutron-server to 500 when maximum sentinel clients are reached on the target sentinel: https://github.com/andymccurdy/redis-py/pull/609

So, now that we've reached agreement, can someone please merge? :)

PR #378, opened 2015-04-28
Implemented get and get all methods for floating IPs

alanquillin:
https://jira.rax.io/browse/NCP-1355
https://jira.rax.io/browse/NCP-1356

- Implemented generic GET methods to the router resource.  These methods return the default generic public router id (all 0's).
- Implemented the GET methods for the floating IPs resource.
- Added framework for the floating IP resources.

*Note: When returning the fixed ip associated with the floating ip, we select the first fixed ip that was allocated for the port.  This is due to the fact that there is no current way to tie a public ip to a specific fixed ip, this needs to be added in a future pull request.*

### DIFF
--- a/quark/plugin_modules/router.py
+++ b/quark/plugin_modules/router.py
@@ -1,0 +1,52 @@
+# Copyright 2013 Openstack Foundation
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from oslo.config import cfg
+from oslo_log import log as logging
+
+CONF = cfg.CONF
+LOG = logging.getLogger(__name__)
+
+quark_router_opts = [
+    cfg.StrOpt('floating_ip_router_id',
+               default='00000000-0000-0000-0000-000000000000',
+               help=_('floating ips default public router id'))
+]
+
+CONF.register_opts(quark_router_opts, "QUARK")
+
+
+def get_router(context, id, fields):
+    LOG.info("get_router for %s for tenant %s fields %s" %
+             (context.tenant_id, fields))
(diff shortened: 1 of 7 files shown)


quark/plugin_views.py
```diff
@@ -40,7 +41,10 @@
     cfg.BoolOpt('show_subnet_ip_policy_id',
                 default=True,
                 help=_('Controls whether or not to show ip_policy_id for'
-                       'subnets'))
+                       'subnets')),
+    cfg.StrOpt('floating_ip_router_id',
+               default='00000000-0000-0000-0000-000000000000',
+               help=_('floating ips default public router id'))
```
asadoughi: Duplicated entry?

asadoughi:
Good stuff here!

I've finished reviewing first draft. Test coverage would be nice, but it looks like the code in this PR might be subject to significant changes since all CRUD operations haven't been implemented yet so I'd be ok with holding off on tests if that's the case.

I'm not sure about all the commented out code. Is that a work in progress (WIP)? Typically, I'd err on the side of not having dead code in the repo, but if it's a WIP thing, I think that's fine.

jmeridth:
:+1: to what Amir said.  I hesitated with no tests being present but saw this is the beginning of a process. :smile:

quark/plugin_views.py
jmeridth: @asadoughi: Duplicate to https://github.com/rackerlabs/quark/pull/378/files#diff-b6bb5573aa8777717e44e17c9537e47dR25 ?

asadoughi: Yes, that's what I was referring to.

alanquillin:
@asadoughi @jmeridth Yes, this is the beginning of the floating IP process so the commented out code is just a WIP progress and will be uncommented once I implement those methods (which I am working on now and should be PRing this week-ish.  I will have a PR for each of the unimplemented methods: delete, update and create).  I do plan on adding a good amount of tests, but wanted to get this framework code in.

alanquillin:
@asadoughi @jmeridth I removed the duplicate conf registration.  @asadoughi Thanks for clarifying that for me!

quark/plugin_modules/router.py
```diff
@@ -0,0 +1,52 @@
...
+               default='00000000-0000-0000-0000-000000000000',
+               help=_('floating ips default public router id'))
+]
+
+CONF.register_opts(quark_router_opts, "QUARK")
+
+
+def get_router(context, id, fields):
+    LOG.info("get_router for %s for tenant %s fields %s" %
+             (context.tenant_id, fields))
```
tr3buchet: bug here I think. too many %s and not enough values in your tuple

PR #379, opened 2015-04-30
Work-in-progress: Shared IPs API RAML specs

asadoughi:
JIRA:NCP-1507


docs/shared_ips.raml
```diff
@@ -0,0 +1,173 @@
...
+                400:
+                    description: |
+                        bad request
+                        * attempting to delete an IP with a status=active port (tenant should disassociate on nova-side first)
+                        * attempting to delete a provider network's IP address
+                404:
+                    description: |
+                        not found
+                        * attempting to delete another tenant's IP address
+        /ports/: # should this URI be hidden from tenants entirely?
```
Cerberus98: Admin-only for sure, less concerned about hidden, but it's probably the best call

docs/shared_ips_nova.raml
```diff
@@ -0,0 +1,87 @@
...
+        description: |
+            List all IPs associated to this instance
+            Invokes and parses (with tenant_id as header): GET http://{neutronUri}/{neutronVersion}/ip_addresses?device_id={serverId}&status=active
+        responses:
+            200:
+                description: ok
+                body:
+                    application/json:
+                        example: |
+                            {"ip_associations": [{"id": "1"}, {"id": "2"}]}
```
Cerberus98: Should we show the actual address as well here? My vote is yes

docs/shared_ips_nova.raml
```diff
@@ -0,0 +1,87 @@
...
+                    description: |
+                        not found
+                        * server not found
+                        * ip_address_id is not found, e.g. neutron call returns 404
+    post:
+        responses:
+            501:
+                description: not implemented
+            # or if we wanted to...
+            # 201: {"ip_association": {"id": $ipAddressId}} but i think idempotent PUT to create makes more sense
```
Cerberus98: Yeah, I like the idea that this is idempotent better

docs/shared_ips.raml
asadoughi: Yeah, this would be more of a repose configuration on the public neutron API side.

docs/shared_ips_nova.raml
asadoughi: OK. That makes sense.

docs/shared_ips.raml
```diff
@@ -0,0 +1,173 @@
...
+    is: [ secured ]
+    get:
+        description: List all IPs for a given tenant
+        queryParameters:
+            device_id:
+                description: filters IP addresses by their ports' device_id
+                type: string
+            status:
+                description: filters IP addresses by their status
+                type: string
```
jkoelker: Should we also allow query by `address_type`?

docs/shared_ips.raml
```diff
@@ -0,0 +1,173 @@
...
+                                "address": "192.168.10.1",
+                                "port_ids": ['2f693cca-7383-45da-8bae-d26b6c2d6718'],
+                                "subnet_id": "f11687e8-ef0d-4207-8e22-c60e737e473b",
+                                "used_by_tenant_id": "2345678",
+                                "version": "4",
+                                "address_type": "fixed"}]}
+    post:
+        description: |
+            Neutron creates an IP address on a specified network (network_id) with a specified IP version (version).
+            A list of device_ids may be optionally specified to create the IP address and added to their respective ports.
```
jkoelker: I would specificly call out that device that have multiple ports *must* use the `port_ids` field, and using the `device_ids` field would be a 400 or 412 or something.

docs/shared_ips.raml
asadoughi: Sure. This list is incomplete, only mentioning the query parameters needed to be implemented for the nova->neutron interaction.

docs/shared_ips.raml
asadoughi: Yep. I'll add an entry under 400 for ambiguous ports.

docs/shared_ips.raml
```diff
@@ -0,0 +1,173 @@
...
+                        application/json:
+                            example: |
+                                {"port": {"id": "dce3dca6-2cba-4e00-95e9-748587f94620", "status": "inactive"}}
+                put:
+                    description: |
+                        Internal call from nova to neutron to update an IP address's status to either "active" or "inactive"
+                    body:
+                        application/json:
+                            example: |
+                                {"port": {"status": "active"}}
```
jkoelker: Api wise, I wonder if we shouldn't have more data here. Like

```json
{"port": {"service": "nova"
             "device_id": nova_uuid}
```

This could then allow us to chain services for things like octavia. It will be creating the ports, and then nova will be using them. So we shouldn't allow them to be deallocated until bolth octavia and nova tell us they are not in use.

docs/shared_ips.raml
```diff
@@ -0,0 +1,173 @@
...
+                            description: |
+                                bad input
+                                * where another port for this given IP is active
+                        401:
+                            description: |
+                                unauthorized, e.g. only nova can make this PUT
+                        404:
+                            description: |
+                                not found, e.g.where this port_id=portId does not exist or belongs to other tenant
+                delete:
```
jkoelker: If we do the above, then we can implement delete to decrement the clock:

```
DELETE /{port_id}?service=nova&device_id={device_id}
```

docs/shared_ips.raml
jkoelker: totes

docs/shared_ips.raml
asadoughi: So, instead of a `status` attribute, the body specifies a `service` attribute. `device_id` is already set when the port is created.

asadoughi: e.g. https://github.com/rackerlabs/quark/blob/master/quark/plugin_modules/ports.py#L101-L103

asadoughi: I guess there's nothing stopping us from updating the port's `device_id`.

docs/shared_ips.raml
asadoughi: I'm not sure what `?service=nova&device_id={device_id}` means if ip_address_id and port_id are already specified.

docs/shared_ips.raml
```diff
@@ -0,0 +1,173 @@
...
+                                    "address_type": "fixed"}
+                404:
+                    description: not found, e.g. another tenant's IP address
+        put:
+            description: update an IP address, e.g. to change ports
+            body:
+                application/json:
+                    example: |
+                        {"ip_address": {
+                            "port_ids": ['275b0516-206f-4421-8e42-1d3d1e4e9fb2', '66811c0a-fdbd-49d4-b1dd-f0f15a329744']}
```
miguellavalle: I assume this wipes out all the previous associations with ports or device_id's. If yes, might be good to clarify it in the description above

docs/shared_ips.raml
jkoelker: Well it assumes that you have multiple services claiming use of the same port. So the example is octavia for the LB's, they will create ports eventually to be containers for IPs for VIPs that they use. Then they instruct nova to use that port. we need to only allow deallocation after bolth nova and octavia (which each have different device_ids, the nova vm id and the lb instance id) have informed us that they are done with it.

docs/shared_ips.raml
asadoughi: Yes. I'm hesitant to call the port to IP relation an association so I don't confuse the end user between a port to IP association and a service to IP association. However, I cannot find a better word than association to describe the relation.

docs/shared_ips.raml
```diff
@@ -0,0 +1,173 @@
...
+                    description: |
+                        bad request
+                        * attempting to delete an IP with a status=active port (tenant should disassociate on nova-side first)
+                        * attempting to delete a provider network's IP address
+                404:
+                    description: |
+                        not found
+                        * attempting to delete another tenant's IP address
+        /ports/: # should this URI be hidden from tenants entirely?
+            get:
```
miguellavalle: I am new to RAML, but it seems to me a description for this get might be useful. Something along the lines of: "List the ports associated with an IP"

docs/shared_ips.raml
asadoughi: Oh okay. So, that sounds like multiple port to IP associations for a given port and IP. Basically, two associations in the instance for nova and octavia. Also, it sounds like migrating the device_id attribute off of the port and into the port to IP association. Hrm, adds a bit of additional complexity.

While this may be a relevant use case in the future, I don't see this as necessary for shared IPs. Particularly, this part of the API is internal to nova/octavia->neutron and can be revised later for future use cases easily.

docs/shared_ips.raml
asadoughi: OK. Will do.

docs/shared_ips.raml
asadoughi: What I will take from this conversation, though, is to rename `status` ("inactive", "active") to `service` ("inactive", "nova") in this revision.

docs/shared_ips_nova.raml
miguellavalle: Nice job! Overall, it makes sense to. Thanks

docs/shared_ips.raml
andyhky: Implementation wise, let's map to a keystone service catalog entry, not an OpenStack service name.

asadoughi:
Closing this PR. Opening new PR with feedback given.

asadoughi:
Hi @miguellavalle I just noticed your comments after uploading a new PR, available here https://github.com/rackerlabs/quark/pull/381 I'll respond to the two existing comments here, but I prefer any new comments to be on the new PR. Thanks.

PR #380, opened 2015-05-01
Used admin context for mac_dealloc

roaet:
JIRA:NCP-1509

Alternative solution is for the tenant to claim the mac address when reallocating.

### DIFF
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -729,17 +729,20 @@ class QuarkIpam(object):
         port["ip_addresses"] = list(
             set(port["ip_addresses"]) - set(ips_removed))
 
+    # NCP-1509(roaet):
+    # - started using admin_context due to tenant not claiming when realloc
     def deallocate_mac_address(self, context, address):
-        mac = db_api.mac_address_find(context, address=address,
+        admin_context = context.elevated()
+        mac = db_api.mac_address_find(admin_context, address=address,
                                       scope=db_api.ONE)
         if not mac:
             raise exceptions.NotFound(
                 message="No MAC address %s found" % netaddr.EUI(address))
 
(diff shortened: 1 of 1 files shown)


asadoughi:
Can you elaborate on the alternative solution? There might be a bugfix for other requests that would be better solved in that way. i.e. if the tenant does not claim a reallocated MAC does the GET /ports still work as expected?

Cerberus98:
I'm confused as to how this broke. The code you've changed here hasn't moved since February of last year. What am I missing?

roaet:
The tenant_id in the DB doesn't change to who currently has it. I don't believe this has any bearing on GET /ports since I am still able to get the port.

Cerberus98:
Oh, because of the mac allocation stuff and tenants

roaet:
https://github.com/roaet/quark/commit/c82fa361b843ecf86b5fa4c57db374cce9f33825

roaet:
There were a lot of changes to mac reallocation or whateverz

roaet:
If we're looking for quick fix due to neutron release I think this is valid. Changing the tenant_id is probably more correct but would take more time.

Cerberus98:
I'm not opposed to this being a quick fix. I don't actually see anything wrong with doing this as admin. I only asked because it wasn't clear to me how it had broken.

asadoughi:
Not blocking this patch, but I'd like to understand better... if it's just tenant_id, then it would be simple to append it here https://github.com/roaet/quark/commit/c82fa361b843ecf86b5fa4c57db374cce9f33825#diff-acddac4b6401f9da48acc821f21c024cR213 But I don't see where tenant_id was being assigned prior to the patch.

asadoughi:
:+1: Merging for now, will settle any confusion later

PR #381, opened 2015-05-04
Shared IPs API RAML specs

asadoughi:
Defined both nova and neutron API specifications.

`git diff asadoughi/shared_ips_raml..asadoughi/shared_ips_docs` to compare to the last PR.

JIRA:NCP-1507

### DIFF
--- a/docs/shared_ips.raml
+++ b/docs/shared_ips.raml
@@ -1,0 +1,195 @@
+#%RAML 0.8
+title: Neutron Shared IPs API
+version: v2.0
+baseUri: https://{neutronUri}/{version}/
+mediaType: application/json
+traits:
+    - secured:
+        headers:
+            X-Auth-Token:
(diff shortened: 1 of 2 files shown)


roaet:
Reading reading.

docs/shared_ips.raml
```diff
@@ -0,0 +1,181 @@
...
+                  example: |
+                      {
+                          "ip_address": {
+                            "network_id": "fda61e0b-a410-49e8-ad3a-64c595618c7e",
+                            "version": 4,
+                            "port_ids": ["6200d533-a42b-4c04-82a1-cc14dbdbf2de"]}
+                      }
+        responses:
+            201:
+                description: created
```
asadoughi: I will be adding an example body here shortly per @miguellavalle's comment on old PR.

PR #382, opened 2015-05-04
Don't lock when selecting v6 subnets

Cerberus98:
NCP-1480

While investigating how to go about incrementing subnets without a lock,
I realized that locking v6 subnets was pointless, as we don't rely on
the next_auto_assign_ip field when generating v6 addresses. While we
don't know what version is available under the ANY IPAM driver, BOTH and
BOTH_REQUIRED explicitly pass versions, allowing us to cut down on the
scenarios where we lock.

This patch is an attempt at a quick-win for mitigating the larger
latency issue, allowing us a bit more time to work on the remaining
tickets.

Please note that the feature itself is toggle-able, to prevent any
fallout if this approach doesn't pan out. It may be enabled by setting
the "ipam_select_subnet_v6_locking" config variable to False under the
[QUARK] config section.

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -597,10 +597,13 @@ def network_delete(context, network):
     context.session.delete(network)
 
 
-def subnet_find_ordered_by_most_full(context, net_id, **filters):
+def subnet_find_ordered_by_most_full(context, net_id, lock_subnets=True,
+                                     **filters):
     count = sql_func.count(models.IPAddress.address).label("count")
     size = (models.Subnet.last_ip - models.Subnet.first_ip)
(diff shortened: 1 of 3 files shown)


Cerberus98:
We might not want to try this and rather wait for me to get a larger, more encompassing patch out the door, but this could be an easy win if we wanted to try something else out. It will really only work when allocation public net v6 IP addresses for reasons explained in the commit message and patch, but it seems to work without locking and I think it's worth a shot. Also as above it can be turned off (and defaults to off) so it shouldn't present any risk

asadoughi:
LGTM.

PR #383, opened 2015-05-07
Update shared_ips.raml

sameersatyam:
Updated attribute names

### DIFF
--- a/docs/shared_ips.raml
+++ b/docs/shared_ips.raml
@@ -35,9 +35,9 @@ traits:
                                 "address": "192.168.10.1",
                                 "port_ids": ['2f693cca-7383-45da-8bae-d26b6c2d6718'],
                                 "subnet_id": "f11687e8-ef0d-4207-8e22-c60e737e473b",
-                                "used_by_tenant_id": "2345678",
+                                "tenant_id": "2345678",
                                 "version": "4",
-                                "address_type": "fixed"}]}
+                                "type": "fixed"}]}
     post:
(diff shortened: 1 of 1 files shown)


jmeridth:
I'm confused.  Thought @asadoughi explained why it needed to be `used_by_tenant_id`.  Are we switching it anyway?

asadoughi:
I asked @sameersatyam to post it here to facilitate discussion. I don't have a strongly held opinion on it being `used_by_tenant_id` as much as @roaet might. To summarize the existing discussion, having it as `tenant_id` would be consistent with other resources. Based on the ownership, in the case of provider networks, `used_by_tenant_id` would make more sense compared to `tenant_id`, but in the case of tenant isolated networks `tenant_id` makes more sense.

jmeridth:
Ah, cool. Thanks for clarifying.

roaet:
I personally don't think this change is necessary. used_by_tenant_id is explicit and is currently how the code works. Should we need to add a tenant_id (for who owns it truly), then we'd have to redo all of whatever we'd be taking out.

jmeridth:
I prefer explicit over implicit as well.  @sameersatyam consistency with other resource that important?

roaet:
I'm fine with this if we change the view. I feel like migrating the DB and losing our explicit attribute is a bad idea.

TMaddox:
:+1:

sameersatyam:
That's good with me and what I was asking all along.

PR #384, opened 2015-05-08
Implemeted the remove floating IP method

alanquillin:
- Implemented the floating ip DELETE api method
- Added a few tests to verify the delete logic
- Fixed an error with the logging of the router api's "get by id" method

### DIFF
--- a/quark/plugin_modules/floating_ips.py
+++ b/quark/plugin_modules/floating_ips.py
@@ -18,7 +18,9 @@ from oslo_log import log as logging
 
 from quark.db import api as db_api
 from quark.db import ip_types
+from quark.drivers import floating_ip_registry as registry
 from quark import exceptions as quark_exceptions
+from quark import ipam
 from quark import plugin_views as v
 
 
@@ -26,9 +28,9 @@ CONF = cfg.CONF
 LOG = logging.getLogger(__name__)
 
 quark_router_opts = [
-    cfg.StrOpt('floating_ip_base_url',
-               default='http://localhost/',
-               help=_('floating ips base url'))
+    cfg.StrOpt('default_floating_ip_driver',
+               default='Unicorn',
+               help=_('Driver for floating IP'))
 ]
 
 CONF.register_opts(quark_router_opts, "QUARK")
@@ -75,7 +77,21 @@ def update_floatingip(context, id, body):
 
 def delete_floatingip(context, id):
     LOG.info("delete_floatingip %s for tenant %s" % (id, context.tenant_id))
-    raise NotImplementedError()
+
+    filters = {"address_type": ip_types.FLOATING, "_deallocated": False}
+
+    addr = db_api.floating_ip_find(context, id=id, scope=db_api.ONE, **filters)
+    if not addr:
+        raise quark_exceptions.FloatingIpNotFound(id=id)
+
+    driver_type = CONF.QUARK.default_floating_ip_driver
+    driver = registry.DRIVER_REGISTRY.get_driver(driver_type)
+
+    driver.remove_floating_ip(addr)
+
+    strategy_name = addr.network["ipam_strategy"]
+    ipam_driver = ipam.IPAM_REGISTRY.get_strategy(strategy_name)
+    ipam_driver.deallocate_ip_address(context, addr)
 
 
 def get_floatingip(context, id, fields=None):
@@ -94,7 +110,7 @@ def get_floatingip(context, id, fields=None):
     """
     LOG.info("get_floatingip %s for tenant %s" % (id, context.tenant_id))
 
-    filters = {"address_type": ip_types.FLOATING}
+    filters = {"address_type": ip_types.FLOATING, "_deallocated": False}
 
     addr = db_api.floating_ip_find(context, id=id, scope=db_api.ONE, **filters)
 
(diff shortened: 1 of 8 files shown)


quark/drivers/registry.py
```diff
@@ -23,7 +24,8 @@ def __init__(self):
         self.drivers = {
             base.BaseDriver.get_name(): base.BaseDriver(),
             optnvp.OptimizedNVPDriver.get_name(): optnvp.OptimizedNVPDriver(),
-            unmanaged.UnmanagedDriver.get_name(): unmanaged.UnmanagedDriver()}
+            unmanaged.UnmanagedDriver.get_name(): unmanaged.UnmanagedDriver(),
+            unicorn.UnicornDriver.get_name(): unicorn.UnicornDriver()}
```
asadoughi: I don't think this belongs here as `UnicornDriver` does not inherit from `BaseDriver`. I'd create your own global singleton for floating IP drivers if that's what you're after.

quark/plugin_modules/floating_ips.py
```diff
@@ -26,9 +28,9 @@
 LOG = logging.getLogger(__name__)
 
 quark_router_opts = [
-    cfg.StrOpt('floating_ip_base_url',
-               default='http://localhost/',
-               help=_('floating ips base url'))
+    cfg.StrOpt('default_floating_ip_driver',
```
asadoughi: Does this represent the default or the actual setting? Consider renaming to `floating_ip_driver`.

Also, on second glance, do you expect to create other backend drivers for floating IP in the immediate future? If not, I would follow the YAGNI principle and remove user-configurable loading all together from this PR.

quark/tests/plugin_modules/test_floating_ips.py
```diff
@@ -0,0 +1,60 @@
...
+class TestRemoveFloatingIPs(test_quark_plugin.TestQuarkPlugin):
+    @contextlib.contextmanager
+    def _stubs(self, addr):
+        addr_model = None
+        if addr:
+            addr_model = models.IPAddress()
+            addr_model.update(addr)
+
+        def _dealloc_ip(context, addr):
+            pass
```
asadoughi: Unnecessary.

quark/tests/plugin_modules/test_floating_ips.py
```diff
@@ -0,0 +1,60 @@
...
+
+        def _remove_flip(addr):
+            pass
```
asadoughi: Also, unnecessary.

quark/tests/plugin_modules/test_floating_ips.py
```diff
@@ -0,0 +1,60 @@
...
+
+        with contextlib.nested(
+            mock.patch("quark.db.api.floating_ip_find"),
+            mock.patch("quark.ipam.QuarkIpam.deallocate_ip_address"),
+            mock.patch("quark.drivers.unicorn_driver.UnicornDriver"
+                       ".remove_floating_ip")
+        ) as (flip_find, mock_dealloc, mock_remove_flip):
+            flip_find.return_value = addr_model
+            mock_dealloc.side_effect = _dealloc_ip
+            mock_remove_flip.side_effect = _remove_flip
```
asadoughi: A mock object will automatically mock out its attributes (and functions) so no need to assign no-op functions.

asadoughi:
I've finished reviewing the first draft. I have a few concerns.

quark/plugin_modules/floating_ips.py
alanquillin: Yes, the plan is that there will potentially be other backend drivers for floating, nit just unicorn but that we will default to unicorn.

quark/drivers/registry.py
alanquillin: The reason it was added as a driver was that the plan is that there will be multiple drivers for floating ips so i followed the current driver factory pattern here instead of creating a new factory for floating ip drivers.

quark/tests/plugin_modules/test_floating_ips.py
alanquillin: Ok cool, thanks, was just being careful, but i will remove them.

alanquillin:
@asadoughi i addressed your feedback for the potential for multiple drivers, do you have other concerns?

quark/drivers/registry.py
asadoughi: Yes, please create a new factory since configuring a floating IP driver as "unmanaged" doesn't make sense, but would be allowable in the current code.

quark/plugin_modules/floating_ips.py
asadoughi: OK, that makes sense then.

So, to the first part of my comment: will there be multiple floating drivers configured at runtime? e.g. a configuration variable for `default_floating_ip_driver` and another one for `floating_ip_driver_number_two`. If not, please rename `default_floating_ip_driver` to `floating_ip_driver`.

asadoughi:
No other concerns then the ones already commented on. Awaiting your second draft.

alanquillin:
@asadoughi I have pushed up the updated code.  I created a new floating ip driver registry and moved the unicorn driver to that.  As for the config value ```default_floating_ip_driver```, I left that the same for the reason that this will act as a fallback when we move to supporting other drivers (like if we ever need to support Juniper).  When we do that, the trigger to tell which driver to use will not come from the config, but instead use an external value (more thank likely by adding to the network record).  If this value is not set, then it will default to this config value.

asadoughi:
Cool. Thanks for helping me understand. The second draft LGTM.

PR #386, opened 2015-05-13
Test setup.cfg change for upstream change

asadoughi:
https://github.com/openstack/neutron/commit/276028cca26af573c14938255e40c58358eabd4a


jmeridth:
:+1:

asadoughi:
This change isn't necessary according to #387 but maybe it will be in the near future.

PR #388, opened 2015-05-18
Implemented the floating ip create api method

alanquillin:
https://jira.rax.io/browse/NCP-1357

- Implemented the create floating ip API method
- Fixed a few inconsistent variable names within the floating ips logic
- Fixed failing migration when using the mysql 5.6.  There is a bug in mysql 5.5 which allows a column to be modified when it is a foreign key ref.  The change temporarily drops the foreign key constraint while modifying the column
- Added new floating ip to fixed ip mapping table
- Fixed an issue where floating ips were being returned via the /ports api methods inside the port's fixed_ip collection


quark/db/migration/alembic/versions/3b467be51e43_composite_primary_key_port_ip_.py
```diff
@@ -50,13 +50,17 @@ def _foreign_keys_dropped(op, table):
...
+    table = sa.Table('quark_port_ip_address_associations', metadata,
+                     autoload=True)
+    with _foreign_keys_dropped(op, table):
+        op.alter_column('quark_port_ip_address_associations', 'ip_address_id',
+                        existing_type=sa.String(36), nullable=False)
+        op.alter_column('quark_port_ip_address_associations', 'port_id',
+                        existing_type=sa.String(36), nullable=False)
+        op.create_primary_key("pk_quark_port_ip_address_associations",
+                              "quark_port_ip_address_associations",
+                              ['port_id', 'ip_address_id'])
```
jmeridth: @asadoughi @alanquillin Do we want to edit pre-existing migrations and not just create another one modifying the existing table?  I understand we haven't gone to production with this yet, but wanted to be explicit about the decision.

quark/db/models.py
```diff
@@ -197,6 +197,40 @@ def formatted(self):
...
+    sa.Column("fixed_ip_address_id", sa.String(36),
+              sa.ForeignKey("quark_ip_addresses.id"), nullable=False,
+              primary_key=True),
+    sa.Column("enabled", sa.Boolean(), default=True, nullable=False,
+              server_default='1'),
+    **TABLE_KWARGS)
+
+orm.mapper(FloatingToFixedIPAssociation, flip_to_fixed_ip_assoc_tbl)
+
+IPAddress.fixed_ip = orm.relationship("IPAddress",
```
jmeridth: why not have this under `class IPAddress` like https://github.com/rackerlabs/quark/blob/master/quark/db/models.py#L166

then a reader gets a one shot view of `IpAddress`

quark/exceptions.py
```diff
@@ -140,5 +140,26 @@ class FloatingIpNotFound(exceptions.NeutronException):
...
+class RegisterFloatingIpFailure(exceptions.NeutronException):
+    message = _("An error occurred when trying to register the floating IP "
+                "%(id).")
+
+
+class PortAlreadyContainsFloatingIp(exceptions.Conflict):
+    message = _("Port %(port_id) already has an associated floating IP.")
+
+
+class FixedIpDoesNotExistsForPort(exceptions.BadRequest):
```
jmeridth: `Exists` or `Exist`?

/me is being pedantic, please ignore if not worth it :smile:

quark/db/models.py
alanquillin: That was my initial thought too, but I tried and could not get it to work.  Something to do with the self referrenciating mapping table.  If you have suggestions on how to get it to work, I would gladly add it to the 'IPAddress' class, as I agree, it would best be there.

quark/db/migration/alembic/versions/3b467be51e43_composite_primary_key_port_ip_.py
alanquillin: I modified the existing one to fix the migration failures when using mysql 5.6.  If we added add this to a new migration, it would still fail when it gets to this one (once we move past mysql 5.5).  I can pull this out if you want.  Wanted to add it to see what you guys thought.

jmeridth: Oh that's right.  :+1:

quark/db/models.py
jmeridth: hrm.  I'll have to give it a try.

PR #389, opened 2015-05-20
Neutron Shared IPs API doc updates

lmaycotte:
- Added query params for the /ip_addresses GET call
- Added query params for the /ip_addresses/{id}/ports GET call
- Added device_id in the GET /ip_addresses/{id}/ports response examples

### DIFF
--- a/docs/shared_ips.raml
+++ b/docs/shared_ips.raml
@@ -16,7 +16,35 @@ traits:
     is: [ secured ]
     get:
         description: List all IPs for a given tenant
-        queryParameters:  # This is not a complete list of query parameters, just the ones relevant to Shared IP.
+        queryParameters:
+            # All from the IP addresses response
+            id:
+                description: filters by IP addresses UUID
+                type: string
+            network_id:
+                description: filters IP addresses by their network UUID
+                type: string
+            address:
+                description: filters IP address by their address
+                type: string
+            # Query will support only one port ID
+            port_ids:
+                description: filters IP addresses by a port UUID
+                type: string
+            subnet_id:
+                description: filters IP addresses by their subnet UUID
+                type: string
+            tenant_id:
+                description: filters IP addresses by their user tenant ID
+                type: string
+            version:
+                description: filters IP addresses by their version 4 or 6
+                type: int
+            type:
+                description: filters IP addresses by their type, for ex. fixed
+                type: string
+            
+            # Plus the following /ip_addresses/{id}/ports child resources
             device_id:
                 description: filters IP addresses by their ports' device_id
                 type: string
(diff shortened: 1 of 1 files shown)


docs/shared_ips.raml
```diff
@@ -16,7 +16,39 @@ traits:
...
+		    	description: filters by IP addresses UUID
+		    	type: string
+		    network_id: 
+		    	description: filters IP addresses by their network UUID
+		    	type: string
+		    address: 
+		    	description: filters IP address by their address
+		    	type: string
+		    # Query with multiple port IDs ex.
+		    # https://{neutronUri}/{version}/ip_addresses?port_ids={port_id1}&port_ids={port_id2}&port_ids{port_idN}
```
asadoughi: I don't think neutron's API stack will let us do this, unfortunately. At most one `port_ids` can be specified IIRC.

docs/shared_ips.raml
```diff
@@ -16,7 +16,39 @@ traits:
     is: [ secured ]
     get:
         description: List all IPs for a given tenant
-        queryParameters:  # This is not a complete list of query parameters, just the ones relevant to Shared IP.
+        queryParameters:
+            # All from the IP addresses response
```
asadoughi: Not a big deal, but the whitespace of this line and following lines seems off. I've been using 4 spaces per indentation level.

docs/shared_ips.raml
```diff
@@ -16,7 +16,39 @@ traits:
...
+		    	type: string
+		    version: 
+		    	description: filters IP addresses by their version 4 or 6
+		    	type: int
+		    type: 
+		    	description: filters IP addresses by their type, for ex. fixed
+		    	type: string            
+            
+            # Plus the /ip_addresses/{id}/ports child resources
+            port_id:
```
asadoughi: `port_ids` is redundant to `port_id` so I think we should stick with just `port_ids`.

asadoughi:
Finished reviewing first draft. I've left 3 comments as feedback. Thanks.

docs/shared_ips.raml
lmaycotte: removing this param from the query list since port_ids will do the exact same.

docs/shared_ips.raml
lmaycotte: updating.

docs/shared_ips.raml
lmaycotte: updating comment mentioning the query will support only one port ID

docs/shared_ips.raml
lmaycotte: Indentation should be fine now.

docs/shared_ips.raml
```diff
@@ -16,7 +16,35 @@ traits:
...
+            network_id:
+                description: filters IP addresses by their network UUID
+                type: string
+            address:
+                description: filters IP address by their address
+                type: string
+            # Query will support only one port ID
+            port_ids:
+                description: filters IP addresses by a port UUID
+                type: string
```
asadoughi: Brandon just showed me that `/networks?name=foo&name=bar` actually works as expected so I'll have to double check how that works in the attribute for an API extension.

asadoughi:
I'm going to look into the multiple same-name query parameter, but this LGTM for now.

PR #390, opened 2015-05-22
Clean up indexes

jmeridth:
Drop all indexes and only add back in those that would be used for
queries.

JIRA:NCP-1530

original author: @jkoelker
https://github.com/jkoelker/quark/commit/7c472764e8b76c5dbb8fc3775f8bcb02db401841


jmeridth:
fixed migration revision SHAs after merging of #388

PR #391, opened 2015-06-05
Implemented the floating ip update API method

alanquillin:
- Implemented the floating IP update logic
- fixed issue with the floating IP create logic where the port was not getting associated with the new floating ip address (in the `quark_port_ip_address_associations` table)
- fixed issue with the floating IP delete logic not cleaning up the port and fixed IP associations

__NOTE:__
The `enable` column of the floating ip to fixed ip assoc table is not used and probably should be removed in a future PR.  This was added under incorrect assumptions/info.  However I did not remove it with the PR as I wanted to have a discussion and its removal is not imperative to the functionality.

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -299,6 +299,12 @@ def ip_address_delete(context, addr):
     context.session.delete(addr)
 
 
+def ip_address_deallocate(context, address, **kwargs):
+    kwargs["_deallocated"] = 1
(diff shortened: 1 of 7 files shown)


asadoughi:
LGTM, but requires rebase.

alanquillin:
@asadoughi Ok, rebase complete.

PR #392, opened 2015-06-08
Shared IPs API

roaet:
Adds support for a new sub resources:

  * /ip_addresses/<id>/ports
  * /ip_addresses/<id>/ports/<id>

While these new resources are essentially the same construct as /ports they
have their own view and thus produce different output.

Current version does not include the code for quotas and policy.

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -85,7 +85,7 @@ def _model_attrs(model):
 def _model_query(context, model, filters, fields=None):
     filters = filters or {}
     model_filters = []
-    eq_filters = ["address", "cidr", "deallocated", "ip_version",
+    eq_filters = ["address", "cidr", "deallocated", "ip_version", "service",
                   "mac_address_range_id", "transaction_id"]
     in_filters = ["device_id", "device_owner", "group_id", "id", "mac_address",
                   "name", "network_id", "segment_id", "subnet_id",
@@ -293,6 +293,10 @@ def ip_address_create(context, **address_dict):
     ip_address["allocated_at"] = timeutils.utcnow()
     context.session.add(ip_address)
     return ip_address
+
+
+def ip_address_delete(context, addr):
+    context.session.delete(addr)
 
 
 @scoped
--- a/quark/db/migration/alembic/versions/4dbf83f37bc0_added_service_column_to_ports_table.py
+++ b/quark/db/migration/alembic/versions/4dbf83f37bc0_added_service_column_to_ports_table.py
@@ -1,0 +1,24 @@
+"""Added service column to port table
+
+Revision ID: 4dbf83f37bc0
+Revises: 33e9e23ba761
+Create Date: 2015-05-26 13:27:38.995202
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '4dbf83f37bc0'
+down_revision = '1bdc1b574beb'
+
+from alembic import op
+import sqlalchemy as sa
+
+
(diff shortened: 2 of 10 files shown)


quark/ip_availability.py
```diff
@@ -47,9 +47,10 @@ def main():
 
 def get_ip_availability(**kwargs):
```
roaet: No idea why this is included. Will remove.

quark/db/models.py
```diff
@@ -144,7 +144,7 @@ class IPAddress(BASEV2, models.HasId):
                                           name="subnet_id_address"),
                       TABLE_KWARGS)
     address_readable = sa.Column(sa.String(128), nullable=False)
-    address = sa.Column(custom_types.INET(), nullable=False, index=True)
```
roaet: Probably accidentally removed.

quark/db/api.py
```diff
@@ -244,7 +244,8 @@ def port_disassociate_ip(context, ports, address):
     return address
 
 
-def port_associate_ip(context, ports, address, enable_port=None):
+def port_associate_ip(context, ports, address, enable_port=None,
```
roaet: Parameter no longer needed. Will be removed.

quark/plugin_modules/ip_addresses.py
```diff
@@ -75,7 +76,15 @@ def validate_port_ip_quotas(context, ports):
 
 def _shared_ip_request(ip_address):
     port_ids = ip_address.get('ip_address', {}).get('port_ids', [])
-    return len(port_ids) > 1
+    device_ids = ip_address.get('ip_address', {}).get('device_ids', [])
```
roaet: effectively makes it so device_ids > 1 or port_ids > 1 == shared ip. If this is not intended then tell me.

quark/db/migration/alembic/versions/4dbf83f37bc0_added_service_column_to_ports_table.py
```diff
@@ -0,0 +1,24 @@
...
+revision = '4dbf83f37bc0'
+down_revision = '1bdc1b574beb'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('quark_ports', sa.Column('service', sa.String(length=255),
+                                           nullable=False,
```
Cerberus98: One more time, why is this explicit "none" and not nullable? Sorry for asking over and over

quark/plugin_modules/ip_addresses.py
Cerberus98: If one IP is on > 1 ports, then it should be shared. Conversely, that 1 IP should *not* be on the same device_id more than once.

We have code in the port creates to prevent you from having a port/network connected to a device more than once, so implicitly any given IP shouldn't be allowed on the same device more that once. Unless I'm missing something here, I think the check is superfluous.

Cerberus98:
I think this accomplishes what we want. Let me know if I'm wrong about that check.

quark/plugin_modules/ip_addresses.py
roaet: This isn't a check to see if it is connected to a device more than once; it is a check to act as the 'flag' for whether or not an IP is shared or not.

Cerberus98: Ahh yep, I misunderstood. Disregard

PR #395, opened 2015-06-09
Delete v6 addresses instead of deallocating

Cerberus98:
NCP-1541

For various reasons, we don't need to keep deallocated v6 address
records in the database. Additionally, with the current indexing plan,
removing them can substantially improve performance of some of our most
important db queries, so we should remove v6 addresses when the port
using them is deleted.


asadoughi:
LGTM

PR #396, opened 2015-06-10
Stop trying to reallocate v6 addresses

Cerberus98:
NCP-1548

After merging NCP-1541, we no longer have v6 addresses to reallocate, so
we should stop trying to reallocate them. This allows us to clean up a
fair amount of code and remove some now redundant tests.

### DIFF
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -516,36 +516,12 @@ class QuarkIpam(object):
                 LOG.info("Generated a new v6 address {0}".format(
                     str(ip_address)))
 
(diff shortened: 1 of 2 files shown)


Cerberus98:
For further testing manually created a case where I marked an allocated MAC as deallocated and then created a new port and it reused that MAC address but successfully fell back to the random bits IP generation, which is exactly what we want.

asadoughi:
:+1:

Cerberus98:
Imma merge my own. I think there's a statute of limitations here

PR #405, opened 2015-06-30
Fixed oslo-based imports

asadoughi:
Changed with openstack/neutron@7863399


asadoughi:
Hmm, http://tarballs.openstack.org/neutron/ appears to be empty.

asadoughi:
tarball host is back up. Looks like there's also an oslo.db to oslo_db change to address

roaet:
LGTM

PR #409, opened 2015-07-01
Fix oslo imports

asadoughi:
sed -i 's/oslo\./oslo_/g' $f

### DIFF
--- a/quark/tests/functional/test_ip_availability.py
+++ b/quark/tests/functional/test_ip_availability.py
@@ -3,11 +3,14 @@ import datetime
 import mock
 import netaddr
 from neutron.db import api as neutron_db_api
-from oslo.config import cfg
+from oslo_config import cfg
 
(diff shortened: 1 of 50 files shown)


quark/tests/functional/test_ip_availability.py
```diff
@@ -31,14 +34,15 @@ def _insert_ip_policy(self, id=0, excludes=None):
             excludes = (0, 255)
         self.connection.execute(
             self.ip_policy.insert(),
-            dict(id=id, size=len(excludes)))
+            id=id, size=len(excludes), created_at=EPOCH)
```
asadoughi: I didn't investigate why this changed under oslo.db to `oslo_db`, but the executor for all models' `created_at` was upset for having not specified a datetime object.

roaet:
LGTM

PR #411, opened 2015-07-17
Pins package versions and fixes unit tests

Cerberus98:
JIRA:NCP-1604


Cerberus98:
fwiw, the versions pinned were what was installed by the previous ranged/unpinned requirements. I realize this isn't the most thorough logic, but all of our tests passed and things HAVE been running against roughly these versions for some time.

jkoelker:
LGTM

PR #416, opened 2015-07-21
Fix IPs landing in wrong subnet with shared IP

roaet:
Fixes JIRA:NCP-1605


Cerberus98:
Changing the method name would be lovely, but otherwise :+1:

PR #419, opened 2015-07-28
Actually including the migrations

roaet:
Fixes JIRA:NCP-1608


asadoughi:
The push test won't go green until it's rebased off of master, but the pr test is green so I'll go ahead and merge.

PR #423, opened 2015-08-03
Fixed quark-db-manage

asadoughi:
No longer able to re-use the upstream neutron-db-manage cli.py because of its branching support, I forked the code here, moving it down one directory to successfully import alembic and not the local alembic directory.


asadoughi:
Please merge #424 first.

jkoelker:
LGTM

PR #426, opened 2015-08-04
Database models and migrations for resource locks

asadoughi:


quark/db/models.py
```diff
@@ -171,6 +171,10 @@ class IPAddress(BASEV2, models.HasId):
     transaction_id = sa.Column(sa.Integer(),
                                sa.ForeignKey("quark_transactions.id"),
                                nullable=True)
+    lock_id = sa.Column(sa.Integer(),
```
jkoelker: Shouldn't the relationship be a many to one so multiple things can lock the same resource?

asadoughi: Rather than implementing multiple locks per resource, quark could have an additional table such that multiple holders could map to the same lock. I'll go ahead and implement it that way if that sounds good. I just couldn't think of an example where there would be multiple holders of the same lock.

jkoelker: :+1: 

the one that comes to mind is octavia and nova bolth using an ip as a vip on the lb and as a fixed ip on the port.

asadoughi:
Closing this PR temporarily to address the suggested changes.

PR #427, opened 2015-08-05
Allow None to be passed in as a filter-value

asadoughi:
Enables finding ip_addresses without locks (lock_id = null).

JIRA:NCP-1622


jkoelker:
LGTM

PR #428, opened 2015-08-05
Allow Overriding the Strategy for FloatingIPs

jkoelker:


asadoughi:
LGTM. However, would it be unusual to ask for the PRs for the configuration changes in advance of merging this PR as it is a breaking change if the configuration were not to be updated?

asadoughi:
Oh, nevermind my last comment. I just realized the change is in `create_floatingip` not `create_port`.

PR #429, opened 2015-08-05
Decouple SG from unmanaged driver

mpath:
Moves the appropriate logic to the sg driver.

JIRA:NCP-1471

### DIFF
--- a/quark/drivers/security_groups.py
+++ b/quark/drivers/security_groups.py
@@ -1,0 +1,46 @@
+# Copyright 2014 Openstack Foundation
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#
+
+from oslo_log import log as logging
+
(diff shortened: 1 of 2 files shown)


quark/drivers/security_groups.py
```diff
@@ -0,0 +1,58 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class SecurityGroupDriver(object):
+
+    def __init__(self):
+        self.is_capable = self.is_security_group_capable()
+
+    def update_port(self, **kwargs):
+        if not self.is_capable:
```
asadoughi: ha

quark/drivers/security_groups.py
```diff
@@ -0,0 +1,58 @@
...
+        client = sg_client.SecurityGroupsClient(use_master=True)
+        try:
+            client.delete_vif(kwargs["device_id"],
+                              kwargs["mac_address"])
+        except Exception:
+            LOG.exception("Failed to reach the security groups backend")
+
+    @env.has_capability(env.Capabilities.SECURITY_GROUPS)
+    def is_security_group_capable(self):
+        return True
```
asadoughi: Memes aside, what's going on here?

mpath: The decorator is being undecorated. Calling the function that looks for a config value every time we call port update and in every function in the sg driver... instead I set a member var to contain the return value on __init__ and use that instead. Same effect. Should save a few cycles over the life of the product.

mpath: Also bike shedding/pet ducking.

asadoughi: I like your intent, but this is not a valuable software engineering practice. See http://c2.com/cgi/wiki?PrematureOptimization for more. Please re-factor with the decorator in place. If the decorator is shown later to be a critical performance issue, we can optimize it then.

mpath: It worked!

PR #430, opened 2015-08-06
Fix tenant ID in get floating IP

andyhky:


alanquillin:
LGTM +1

Cerberus98:
I would have done this completely differently but I guess this is fine

PR #431, opened 2015-08-12
Isolated networks security groups

mpath:
Removes security groups from NVP drivers and uses
security groups driver instead.

JIRA:NCP-1465


quark/drivers/nvp_driver.py
```diff
@@ -304,16 +301,15 @@ def _create_lswitch_port():
         return _create_lswitch_port()
 
     @utils.retry_loop(CONF.NVP.operation_retries)
-    def update_port(self, context, port_id, status=True,
-                    security_groups=None, **kwargs):
-        security_groups = security_groups or []
+    def update_port(self, context, port_id, mac_address, device_id,
```
asadoughi: `mac_address` and `device_id` should have default values if they are being treated as kwargs.

asadoughi:
Finished reviewing first draft. Just one suggested change. Thanks.

PR #433, opened 2015-08-25
Exception messages now passed through

roaet:
Fixes JIRA:NCP-1592

Might fix JIRA:NCP-1631, JIRA:NCP-1630


Cerberus98:
:snowman: :egg: :hankey: :dog2:

PR #435, opened 2015-08-26
Prevent delete fixed_ips & remove tenant checks

roaet:
Fixes JIRA:NCP-1630

The tenant checks were removed from the explicit queries because they are added
to the queries in _model_query in db.api.

### DIFF
--- a/quark/plugin_modules/ip_addresses.py
+++ b/quark/plugin_modules/ip_addresses.py
@@ -211,7 +211,6 @@ def create_ip_address(context, body):
     segment_id = validate_and_fetch_segment(ports, network_id)
     if iptype == ip_types.SHARED:
         old_addresses = db_api.ip_address_find(context,
-                                               tenant_id=context.tenant_id,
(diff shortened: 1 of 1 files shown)


quark/plugin_modules/ip_addresses.py
```diff
@@ -327,10 +326,16 @@ def delete_ip_address(context, id):
     LOG.info("delete_ip_address %s for tenant %s" % (id, context.tenant_id))
     with context.session.begin():
         ip_address = db_api.ip_address_find(
-            context, id=id, tenant_id=context.tenant_id, scope=db_api.ONE)
+            context, id=id, scope=db_api.ONE)
```
Cerberus98: So since this removes the tenant check, technically you'd find an IP that wasn't yours. I know Neutron likes to filter those out in the API layer. What response do you actually get?

Cerberus98:
:eggplant: :cactus:

PR #436, opened 2015-08-31
Adding VLAN ID to Port objects via Tags

ClifHouck:
Adds a module to store and retrieve VLAN IDs using Tags on the Port
model. Also, VLAN ID data is attached to Port descriptions, where
appropriate,  obtained via port show API requests.

Along with necessary unit tests.


Cerberus98:
Crap, just realized this one is still here. Sorry @ClifHouck

asadoughi:
Is this still a "work in progress", though? I routinely ignored it because of the title.

quark/plugin_views.py
```diff
@@ -179,6 +180,11 @@ def _port_dict(port, fields=None):
     # NOTE(mdietz): more pythonic key in dict check fails here. Leave as get
     if port.get("bridge"):
         res["bridge"] = port["bridge"]
+
+    vlan_id = port_vlan_id.retrieve_vlan_id(port)
+    if vlan_id:
+        res["vlan_id"] = vlan_id
+
```
Cerberus98: Since tags aren't eager loaded, this will result in another trip in the view to the database. It's trivial so it shouldn't be an issue, but if I find anything else I'll ask you to respin this with a TODO or NOTE on that, saying we may want to eager load it some time down the road.

quark/port_vlan_id.py
```diff
@@ -0,0 +1,121 @@
...
+    is_vlan_id_tag(tag).
+
+    :param tag: The tag object.
+    :returns: The VLAN ID as an integer if extraction is successful
+       Otherwise returns None.
+    """
+    try:
+        vlan_id = int(tag[len(VLAN_TAG_PREFIX):])
+    except Exception:
+        return None
```
Cerberus98: How often is this going to fail? Exceptions will pause all concurrency in the eventlet hub while they're resolved, so if this was going to happen most of the time, I think we'd want to take an alternate route.

I wouldn't be harping on performance so much except it's been my focus lately to actually do something about it for a change :-)

Cerberus98:
@asadoughi Yeah, I think it's still WIP, I remember getting asked to look at it. It's up because he wants feedback

jimrollenhagen:
@Cerberus98 This is only WIP in the sense that we haven't actually tried to run it yet :) assuming it works, there's no plan to change the code.

quark/port_vlan_id.py
ClifHouck: It should almost never fail, since there's a _validate_vlan_id call made before the system stores VLAN_IDs. The int() call can potentially raise an Exception, so I figured I should handle it.

ClifHouck:
What @jimrollenhagen said.

ClifHouck:
@asadoughi, @Cerberus98 I should change the title, it's ready for use. I put it up so @morgabra can pull it down and use it before it gets merged with quark.

quark/plugin_views.py
ClifHouck: Added a note for you.

PR #437, opened 2015-09-02
(1 of 5) Fix quark-db-manage from upstream change

asadoughi:
neutron.common.repos module was removed

JIRA:NCP-1664


Cerberus98:
:+1

PR #439, opened 2015-09-10
(1 of 4) Added db-level api methods to create, delete locks/lock-holders

asadoughi:
JIRA:NCP-1620

See #437 #438 first.


mpath:
Can we not name it around "Locks"? That implies database or thread locking, would "contract" or "claim" be more appropriate?

asadoughi:
I've rebased the PR to show the actual diff. The naming was already merged in PR #438.

Cerberus98:
:+1:

PR #440, opened 2015-09-14
(2 of 4) IPAM: do not reallocate locked IPs

asadoughi:


quark/ipam.py
```diff
@@ -367,6 +367,7 @@ def attempt_to_reallocate_ip(self, context, net_id, port_id, reuse_after,
             "network_id": net_id,
             "deallocated": True,
             "version": version,
+            "lock_id": None,
```
Cerberus98: We'll probably want to do redo our indexes to include lock_id

Cerberus98:
:+1:

PR #442, opened 2015-09-16
Move to SQLAlchemy 1.0.8 and drop our hack

Cerberus98:
Re-pins SQLAlchemy at 1.0.8 in requirements.txt, which allows us to
remove our hack we were maintaining before the same change merged
upstream. Removing code! Huzzah!


Cerberus98:
Github weirdly merged these into the same PR

asadoughi:
This needs to be separated into 2 PRs.

Cerberus98:
Cranky today, huh

PR #448, opened 2015-09-24
Updates alembic to 0.8.2 to match neutron

Cerberus98:


Cerberus98:
Needed to un-hose our neutron merge ply job

asadoughi:
:+1: pending tests

PR #450, opened 2015-09-25
(4 of 4) Tool to manage locks on null-routed addresses

asadoughi:
Creates and deletes lock holders for null-routed IP addresses.

JIRA:NCP-1624

### DIFF
--- a/quark/tools/null_routes.py
+++ b/quark/tools/null_routes.py
@@ -1,0 +1,187 @@
+import sys
+
+import netaddr
+from neutron.common import config
+from neutron import context as neutron_context
+from oslo_config import cfg
+from oslo_log import log as logging
+import requests
+from sqlalchemy import not_
+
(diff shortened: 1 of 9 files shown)


quark/tools/null_routes.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # NOTE(asadoughi): assertions to ensure schema hasn't changed
+    assert len(data) == 1
+    assert sorted(data[0].keys()) == sorted([
+        "paginate", "request", "payload", "response"])
+    assert sorted(data[0]["paginate"].keys()) == sorted([
+        "total_count", "total_count_display", "total_pages",
+        "author_comment", "per_page", "page"])
+    # NOTE(asadoughi): API response claims to be paged, but all the data is on
+    #                  the first page.
```
Cerberus98: huehuehue

PR #454, opened 2015-10-06
Isolated security groups with conf toggle

mpath:
Allow isolated networks security groups using the
QUARK.environment_capabilities configuration value.

JIRA:NCP-1465
JIRA:NCP-1681

### DIFF
--- a/quark/drivers/nvp_driver.py
+++ b/quark/drivers/nvp_driver.py
@@ -26,6 +26,8 @@ from oslo_config import cfg
 from oslo_log import log as logging
 
 from quark.drivers import base
+from quark.drivers import security_groups as sg_driver
+from quark.environment import Capabilities
 from quark import exceptions
 from quark import utils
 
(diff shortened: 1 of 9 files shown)


quark/drivers/nvp_driver.py
```diff
@@ -304,16 +309,24 @@ def _create_lswitch_port():
         return _create_lswitch_port()
 
     @utils.retry_loop(CONF.NVP.operation_retries)
-    def update_port(self, context, port_id, status=True,
-                    security_groups=None, **kwargs):
-        security_groups = security_groups or []
+    def update_port(self, context, port_id, mac_address=None, device_id=None,
+                    status=True, security_groups=None, **kwargs):
+        if not self.sg_driver:
```
Cerberus98: This bit is confusing to me. Why do we need to set them up here if we aren't going to do anything with them down below? Subsequent usages of "security_groups" are all gated by "if self.sg_driver"

Cerberus98:
Seems alright. Just curious about that one bit :+1:

PR #455, opened 2015-10-06
Fix security groups egress toggle

mpath:
Originally branched on the egress toggle too late in the process,
so we were never blocking creating egress rules.

JIRA:NCP-1681


Cerberus98:
Looks like this needs the previous patch to merge, first

Cerberus98:
Looks ok otherwise :+1:

PR #457, opened 2015-10-06
Updated the floating ip POST/PUT unicorn request to adhere to the new structure

alanquillin:
also fixed the acceptable response codes that it will receive from unicorn (either 200 or 201).

also, ran and verified that the unit tests pass

### DIFF
--- a/quark/drivers/unicorn_driver.py
+++ b/quark/drivers/unicorn_driver.py
@@ -18,12 +18,12 @@ Unicorn driver for Quark
 """
 
 import json
-import netaddr
 import requests
 
 from oslo_config import cfg
 from oslo_log import log as logging
 
+from quark.db import ip_types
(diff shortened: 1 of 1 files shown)


quark/drivers/unicorn_driver.py
```diff
@@ -53,7 +53,7 @@ def register_floating_ip(self, floating_ip, port, fixed_ip):
         LOG.info("Calling unicorn to register floating ip: %s %s" % (url, req))
         r = requests.post(url, data=json.dumps(req))
 
-        if r.status_code != 200:
+        if r.status_code != 200 and r.status_code != 201:
```
Cerberus98: you could also do:

if r.status_code not in [200, 201]

Cerberus98:
:+1:

PR #462, opened 2015-10-09
Creamy German Style

Cerberus98:


jkoelker:
LGTM

evanscottgray:
wat

asadoughi:
@evanscottgray https://en.wikipedia.org/wiki/Quark_(dairy_product) compared to https://en.wikipedia.org/wiki/Quark_(Star_Trek) or https://en.wikipedia.org/wiki/Quark

PR #464, opened 2015-10-09
Remove unused network_strategy code

Cerberus98:
JIRA:NCP-1580

Children of parent networks was a feature that went unused in quark,
and may as well be removed. Additionally, it changes the word "parent"
to be "provider" whereever this code is used. This patch acts as a
pre-cursor to the provider subnets implementation.


Cerberus98:
Repushed for dumb spelling

asadoughi:
:+1:

PR #465, opened 2015-10-12
Revert "WIP: Adding VLAN ID to Port objects via Tags"

Cerberus98:
This reverts commit 8fb86aa58f188de0dcbece33f68fec5abf40182a.

Unfortunately, the original PR is missing a database migration to
actually add the tag_id to quark_ports. This means that the tag UUID
won't be present on quark ports, which breaks all port interactions.


Cerberus98:
I'm told Clif might actually have a fix for this, so don't merge yet

Cerberus98:
https://github.com/rackerlabs/quark/pull/467

PR #468, opened 2015-10-13
Added check for tenant_id

roaet:
JIRA:NCP-1729

This change is hard to test because removing tenant_id from the context
causes all of the subordinate creates to fail. This will cover many 500
errors that we weren't tracking and convert them to appropriate 400s.

### DIFF
--- a/quark/plugin.py
+++ b/quark/plugin.py
@@ -23,6 +23,7 @@ from neutron.quota import resource as qres
 from neutron.quota import resource_registry as qres_reg
 from oslo_config import cfg
 from oslo_log import log as logging
+import webob.exc
 
 from quark.api import extensions
 from quark import ip_availability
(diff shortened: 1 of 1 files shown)


Cerberus98:
:+1:

asadoughi:
:-1: not source of bug

roaet:
@asadoughi this is still a necessary change.

asadoughi:
That's true. The question I'm debating now is, does the error message make sense in the context of a public neutron API endpoint? I'm thinking in the case the identity middleware fails a 400 is better than a 500 so yes.

roaet:
This is common code between the public endpoint and internal endpoint. The tenant_id SHOULD NOT be missing in any public endpoint situation unless someone is bypassing repose.

If they are bypassing repose they are essentially using neutron in noauth configuration and this error message is the appropriate upstream one. If they are using an internal endpoint this error message is the expected message.

PR #471, opened 2015-10-14
Fixed issue where floating ips were not cleaned up correctly on port delete

alanquillin:
Fixes JIRA:NCP-1724
Fixes JIRA:NDP-749

This modifies the port delete process to check ip address before deallocating to see if it is a floating ip.  If it is a floating ip then it does not need to be deallocated, but the fixed ip does need to be disassociated.


asadoughi:
:+1:

PR #473, opened 2015-10-20
WIP: Add generic tag registry for tagged models

morgabra:
Additionally, move vlan_id for ports to this registry.

This provides a easy to extend and generic way to access arbitrary key/value pairs using the existing model tag system. A sort of rudimentary metadata store.

The first use of this will be storing vlan_ids on the Port model if required by the backend driver, and is implemented here as an example. There currently exists no way to actually set any tags yet, but will come in future PRs.

This is pretty defensive, and shouldn't be affected by any existing tags and won't throw when looking up tag data for Ports. We can tighten this up in the future if desired.

All tests included.


morgabra:
TODO: fix docstrings

morgabra:
TODO: Tags don't cascade delete when their parent is deleted, this is probably wrong?

quark/tags.py
```diff
@@ -0,0 +1,169 @@
...
+            vlan_id_int = int(value)
+            assert vlan_id_int >= self.MIN_VLAN_ID
+            assert vlan_id_int <= self.MAX_VLAN_ID
+        except Exception:
+            msg = ("Invalid VLAN ID detected. Got '%(vlan_id)s'. "
+                   "VLAN ID should be an integer between %(min)d and %(max)d "
+                   "inclusive." % {'vlan_id': value,
+                                   'min': self.MIN_VLAN_ID,
+                                   'max': self.MAX_VLAN_ID})
+            raise TagValidationError(value, msg)
```
Cerberus98: Same question I had last time I saw this: This will raise a 500 if someone passes a bad vlan ID, won't it? Is that the servers fault?

Cerberus98:
Just the one thing, but otherwise seems good

quark/tags.py
morgabra: Hmm, yeah. Users can't set this, but it's possible we introduce user-modifiable tags in the future I guess.

Should I raise BadRequest here? Or catch it in db_api when tags are actually set on the model and re-raise it?

Cerberus98: I'd be fine with a BadRequest. Primary reason I'm jumping on it is 500s turn into things I have to investigate later because people want "due diligence" :-P

quark/tags.py
```diff
@@ -0,0 +1,169 @@
...
+    tags = {}
+
+    def get_all(self, model):
+        """Get all known tags from a model.
+
+        Returns a dict of {<tag_name>:<tag_value>}.
+        """
+        tags = {}
+        for name, tag in self.tags.items():
+            for mtag in model.tags:
```
asadoughi: It would be nice to be able to go from a `model.tags` element to a `Tag.NAME` so you could do a `O(1)` lookup instead of `O(N^2)`. This is not a problem right now with only 1 tag of course, but something to watch out for in the future.

morgabra:
I think I broke this pr - I made the changes and rebased + force-pushed to my branch. I'll figure out if I can fix this or not.

morgabra:
bleh, annoying. I'll just open another one.

morgabra:
I messed this one up, pushed to the wrong remote, so no wonder it didn't show up :/

Anyway, latest is #476

PR #474, opened 2015-10-20
add port-level network_plugin

morgabra:
Currently, the driver associated with a Network() (via the network_plugin attribute) determines which driver is used during Port() create/update/deletes on that network.

It is useful to override this behavior, so this adds 'network_plugin' to the port model, and will use that driver instead of the network driver if specified and explicitly allowed.


morgabra:
TODO: This currently will fallback to the net driver if the port driver is not specified. (Which can happen if you make a call outside an admin context also)

It might be better to just throw early if this field is included in the request but you aren't an admin, because if you are including it it's probably for a reason, and you'll end up with a broken port anyway if you use the wrong driver.

Cerberus98:
:+1:

asadoughi:
LGTM. Is this PR still a work in progress?

morgabra:
Nope, this should be good to go.

PR #477, opened 2015-10-23
Implements provider subnets

Cerberus98:
JIRA:NCP-1580

Implements hooks for showing provider subnets, which are merely aliases
for subnets owned by provider networks. Unlike provider networks, they
are merely placeholders and will never be related in the schema to any
other object.

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -566,7 +566,7 @@ def network_find(context, limit=None, sorts=None, marker=None,
     defaults = []
     provider_query = False
     if "id" in filters:
-        ids, defaults = STRATEGY.split_network_ids(context, filters["id"])
+        ids, defaults = STRATEGY.split_network_ids(filters["id"])
         if not ids and defaults and "shared" not in filters:
             provider_query = True
         if ids:
@@ -575,7 +575,7 @@ def network_find(context, limit=None, sorts=None, marker=None,
             filters.pop("id")
 
     if "shared" in filters:
-        defaults = STRATEGY.get_assignable_networks(context)
+        defaults = STRATEGY.get_provider_networks()
         if True in filters["shared"]:
             if ids:
                 defaults = [net for net in ids if net in defaults]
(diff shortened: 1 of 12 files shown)


Cerberus98:
It's a fair amount of change but most of it is copy-pasta when you get down to it. However, this also necessitates a config change

Cerberus98:
Finally, I have a script to go and create the provider subnets we need that would have to be run against each of our production databases. https://gist.github.com/Cerberus98/5f163e44fc0430d6a515

asadoughi:
I think the gist should be a migration a la https://github.com/rackerlabs/quark/blob/master/quark/db/migration/alembic/versions/28e55acaf366_populate_ip_policy_size.py

Cerberus98:
@asadoughi I started to add it as a migration but then I had the thought that one could argue it's Business Specific™ and therefore if there were other hypothetical consumers of quark they wouldn't want our provider config, especially since we didn't add our provider networks as migration

quark/db/api.py
```diff
@@ -231,7 +231,7 @@ def port_count_all(context, **filters):
 
 
 def port_create(context, **port_dict):
-    port = models.Port(tags=[])
```
asadoughi: This looks like you'll need to rebase to un-break vlan tags.

asadoughi:
@Cerberus98 ok but it should be part of the code somewhere so it doesn't get lost. Perhaps only on the internal release branch.

Cerberus98:
@asadoughi yeah that seems fine to me. I considered making an internal migration but I couldn't think of a way to keep that in sync without also having a dummy migration with the same number on public github

quark/db/api.py
```diff
@@ -708,20 +708,59 @@ def subnet_update_set_alloc_pool_cache(context, subnet, cache_data=None):
...
+            defaults.pop(0)
+        if filters and invert_defaults:
+            query = query.filter(and_(not_(models.Subnet.id.in_(defaults)),
+                                      and_(*model_filters)))
+        elif not provider_query and filters and not invert_defaults:
+            query = query.filter(or_(models.Network.id.in_(defaults),
+                                     and_(*model_filters)))
 
-    if "join_dns" in filters:
-        query = query.options(orm.joinedload(models.Subnet.dns_nameservers))
```
asadoughi: Uh, this and the `"join_routes"` if statement were removed?

asadoughi:
Can you detail what the config change would look like, please?

Cerberus98:
@asadoughi oh sorry, thought I'd pasted it already

default_net_strategy='{"00000000-0000-0000-0000-000000000000": {"bridge": "xenbr0", "subnets": ["00000000-0000-0000-0000-000000000000", "11111111-1111-1111-1111-111111111111"]}, "11111111-1111-1111-1111-111111111111": {"bridge": "xenbr1", "subnets": ["22222222-2222-2222-2222-222222222222", "33333333-3333-3333-3333-333333333333"]}}'

Adds "subnets" keys which have lists of values that are just the "ids" of the provider subnets

quark/db/api.py
Cerberus98: ahh yeah, good catch

I'll add those back in shortly

quark/db/api.py
Cerberus98: I see what happened. git preferred the wrong side of the merge for some reason. This line didn't conflict :-/

quark/db/api.py
```diff
@@ -708,20 +708,59 @@ def subnet_update_set_alloc_pool_cache(context, subnet, cache_data=None):
 
 
 @scoped
-def subnet_find(context, limit=None, page_reverse=False, sorts=None,
-                marker_obj=None, **filters):
-    if "shared" in filters and True in filters["shared"]:
-        return []
+def subnet_find(context, limit=None, sorts=None, marker=None,
+                page_reverse=False, fields=None, **filters):
```
asadoughi: kwarg list got re-arranged?

quark/db/api.py
```diff
@@ -708,20 +708,59 @@ def subnet_update_set_alloc_pool_cache(context, subnet, cache_data=None):
...
+            query = query.filter(and_(not_(models.Subnet.id.in_(defaults)),
+                                      and_(*model_filters)))
+        elif not provider_query and filters and not invert_defaults:
+            query = query.filter(or_(models.Network.id.in_(defaults),
+                                     and_(*model_filters)))
 
-    if "join_dns" in filters:
-        query = query.options(orm.joinedload(models.Subnet.dns_nameservers))
+        elif not invert_defaults:
+            query = query.filter(models.Subnet.id.in_(defaults))
```
asadoughi: Why are the filters not applied in this statement?

Cerberus98: For both of these, it's because I copied the existing "network_find" implementation and simply changed out all the instances of "network" with "subnet"

This one for sure is a good catch, and implies network_find is busted too.

asadoughi:
:+1: will merge Monday morning if no one else has objections

PR #481, opened 2015-10-27
Added bridge to backend port filter list

asadoughi:
JIRA:NCP-1743


asadoughi:
Need https://github.com/rackerlabs/quark/pull/480 merged first to fix random UT failure.

PR #482, opened 2015-10-27
Fixed typo

asadoughi:
Typo led to looking up all subnets on provider networks, which is
not expected when only provider subnets are requested.

JIRA:NCP-1580


asadoughi:
Requires #480 to merge first.

PR #484, opened 2015-10-28
Exclude provider subnets from overlapping CIDR validation

asadoughi:
Including the provider subnets as part of the validation caused
a BadRequest as their cidrs span the entire IP space. This would
prevent quark administrators creating any new subnets under provider
networks.

JIRA:NCP-1746


asadoughi:
Requires #480 to merge first.

PR #485, opened 2015-10-29
Exclude provider subnets from query

asadoughi:
Provider subnets' CIDRs span the entire IP space, so including them
causes issues for null_routes. We exclude them by setting shared
to False on the subnet_find as well as reloading python modules
so they can use the appropriately initialized default_net_strategy
configuration parameter.

JIRA:NCP-1749


asadoughi:
Requires #480

PR #487, opened 2015-11-24
Modifies responses to return provider subnet IDs

Cerberus98:
JIRA:NCP-1580

Updates the port view functionality to replace the real subnet_id with a
provider one. Necessitates updating the config again to incorporate the
ip_version in the provider subnet variable.


Cerberus98:
Not *super* thrilled with this implementation but I'm not sure what else to suggest. I'll make the neutron-config change if people are ok with this.

asadoughi:
:+1: LGTM

PR #492, opened 2015-12-17
Fix missing mac range exception

mpath:
If the mac range is missing, which should never happen mind you,
go ahead and delete anyway.

JIRA:NCP-1770


asadoughi:
:+1:

PR #497, opened 2016-01-14
Update API Extension to use

andyhky:
In https://github.com/openstack/neutron/commit/4595899f7f2b3774dc2dac2f8dd1a085b1e7973d
Neutron's extension loading changed. Update extensions that would be ignored.

JIRA:NCP-1782


mpath:
:+1:

PR #500, opened 2016-01-23
Proxy backend driver

morgabra:
This is unfortunately named, because really, it's a pure http proxy backend driver. I'm open to renaming everything to reflect that if you like. 

The idea here is to let quark be in charge of allocation of macs/ips, validating quotas, access control, lifecycle of the port, etc. Instead of calling to NSX or whatever on port create, it instead hits an arbitrary HTTP endpoint with the collected port data.

This supports an agent-like pattern, where the agent is free to manage the actual configuration necessary. This is pretty useful in a bare metal context, given that we have to tightly manage the lifecycle of the actual live configuration on the host, and most of that process is pretty unique to the service and didn't seem like it fit trying to shoehorn the whole thing into quark proper, or at least not at first.

This also adds an IPAM that is suitable for allocating single ips from groups of very small subnets, instead of multiple ips out of larger ones. (Think a bunch of /30s or something)

### DIFF
--- a/quark/api/extensions/ports_quark.py
+++ b/quark/api/extensions/ports_quark.py
@@ -49,7 +49,11 @@ EXTENDED_ATTRIBUTES_2_0 = {
             "is_visible": True},
         "network_plugin": {"allow_post": True, "allow_put": False,
                            "enforce_policy": True,
-                           "is_visible": False, "default": ''}}}
+                           "is_visible": False, "default": ''},
+        "instance_node_id": {"allow_post": True, "allow_put": False,
+                             "default": '', "is_visible": False},
+    }
+}
 
 
 class Ports_quark(extensions.ExtensionDescriptor):
--- a/quark/drivers/ironic_driver.py
+++ b/quark/drivers/ironic_driver.py
@@ -1,0 +1,542 @@
+# Copyright 2013 Openstack Foundation
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+"""
+Ironic agent driver for Quark.
+"""
+import netaddr
+
+from neutron.common import exceptions
+
+from oslo_config import cfg
+from oslo_log import log as logging
+from oslo_utils import importutils
+
+from quark.drivers import base
+from quark import utils
+
+import json
+
+from quark import network_strategy
+
+STRATEGY = network_strategy.STRATEGY
+
+LOG = logging.getLogger(__name__)
+
+CONF = cfg.CONF
+
+IRONIC_IPAM_STRATEGIES = {
+    "provider": {
+        "default": "IRONIC_ANY",
+        "overrides": {
+            "BOTH_REQUIRED": "IRONIC_BOTH_REQUIRED",
+            "BOTH": "IRONIC_BOTH",
+            "ANY": "IRONIC_ANY"
+        }
+    },
+    "tenant": {
+        "overrides": {}
+    }
+}
+
+ironic_opts = [
+    # client connection info
+    cfg.StrOpt("ironic_client", default="neutronclient.v2_0.client.Client",
+               help=("Class of the client used to communicate with the "
+                     "Ironic agent")),
+    cfg.StrOpt('endpoint_url',
+               default='http://127.0.0.1:9697',
+               help='URL for connecting to neutron'),
+    cfg.IntOpt('timeout',
+               default=30,
+               help='Timeout value for connecting to neutron in seconds.'),
+    cfg.BoolOpt('insecure',
+                default=False,
+                help='If set, ignore any SSL validation issues'),
+    cfg.StrOpt('ca_cert',
+               help=('Location of CA certificates file to use for '
+                     'neutron client requests.')),
+    cfg.StrOpt('password',
+               help='Password for connecting to neutron.', secret=True),
+    cfg.StrOpt('tenant_name',
+               help='Tenant name for connecting to neutron.'),
+    cfg.StrOpt('tenant_id',
+               help='Tenant id for connecting to neutron.'),
+    cfg.StrOpt('auth_url',
+               default='http://localhost:5000/v2.0',
+               help='Authorization URL for connecting to neutron.'),
+    cfg.StrOpt('auth_strategy',
+               default='keystone',
+               help='Authorization strategy for connecting to neutron.'),
+
+    # client retry options
+    cfg.IntOpt("operation_retries",
+               default=3,
+               help="Number of times to attempt client operations."),
+    cfg.IntOpt("operation_delay",
+               default=0,
+               help="Base seconds to wait between client attempts."),
+    cfg.IntOpt("operation_backoff",
+               default=0,
+               help="Base seconds for client exponential backoff"),
+
+    # conf options
+    cfg.StrOpt("ironic_ipam_strategies",
+               default=json.dumps(IRONIC_IPAM_STRATEGIES),
+               help="Default IPAM strategies and overrides for this driver")
+]
+
+CONF.register_opts(ironic_opts, "IRONIC")
+
+
+class FakeIronicClient(object):
+    def __init__(self, *args, **kwargs):
+        pass
+
+    def create_port(self, *args, **kwargs):
+        return {"port": {"vlan_id": 500, "id": "fake_uuid"}}
+
+    def delete_port(self, *args, **kwargs):
+        return
+
+
+class IronicException(exceptions.NeutronException):
+    message = "ironic driver error: %(msg)s"
+
+
+class IronicDriver(base.BaseDriver):
+    def __init__(self):
+        self._client = None
+        self._ipam_strategies = None
+        super(IronicDriver, self).__init__()
+
+    @classmethod
+    def get_name(cls):
+        return "IRONIC"
+
+    def load_config(self):
+        LOG.info("Loading Ironic settings.")
+        self._client_cls = importutils.import_class(CONF.IRONIC.ironic_client)
+        self._client = self._get_client()
+        self._ipam_strategies = self._parse_ironic_ipam_strategies()
+        LOG.info("Ironic Driver config loaded. Client: %s"
+                 % (CONF.IRONIC.endpoint_url))
+
+    def _get_client(self):
+
+        params = {
+            'endpoint_url': CONF.IRONIC.endpoint_url,
+            'timeout': CONF.IRONIC.timeout,
+            'insecure': CONF.IRONIC.insecure,
+            'ca_cert': CONF.IRONIC.ca_cert,
+            'auth_strategy': CONF.IRONIC.auth_strategy,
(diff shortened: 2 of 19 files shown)


roaet:
Will most likely merge this shortly. Leaving it up so I can use it for educational purposes with quade and alexm.

quark/api/extensions/ports_quark.py
```diff
@@ -49,7 +49,11 @@
             "is_visible": True},
         "network_plugin": {"allow_post": True, "allow_put": False,
                            "enforce_policy": True,
-                           "is_visible": False, "default": ''}}}
+                           "is_visible": False, "default": ''},
+        "instance_node_id": {"allow_post": True, "allow_put": False,
```
roaet: Is there a situation where a non-admin user can put this in the body of the POST disregarding onmetal-ness?

quark/drivers/ironic_driver.py
```diff
@@ -0,0 +1,542 @@
...
+    }
+}
+
+ironic_opts = [
+    # client connection info
+    cfg.StrOpt("ironic_client", default="neutronclient.v2_0.client.Client",
+               help=("Class of the client used to communicate with the "
+                     "Ironic agent")),
+    cfg.StrOpt('endpoint_url',
+               default='http://127.0.0.1:9697',
```
roaet: Just curious about this default.

quark/drivers/ironic_driver.py
```diff
@@ -0,0 +1,542 @@
...
+    cfg.StrOpt('tenant_name',
+               help='Tenant name for connecting to neutron.'),
+    cfg.StrOpt('tenant_id',
+               help='Tenant id for connecting to neutron.'),
+    cfg.StrOpt('auth_url',
+               default='http://localhost:5000/v2.0',
+               help='Authorization URL for connecting to neutron.'),
+    cfg.StrOpt('auth_strategy',
+               default='keystone',
+               help='Authorization strategy for connecting to neutron.'),
```
roaet: I assume the neutron+quark -> neutron+ironic interaction will be noauth?

quark/drivers/ironic_driver.py
```diff
@@ -0,0 +1,542 @@
...
+
+    def select_ipam_strategy(self, network_id, network_strategy, **kwargs):
+        """Return relevant IPAM strategy name.
+
+        :param network_id: neutron network id.
+        :param network_strategy: default strategy for the network.
+
+        NOTE(morgabra) This feels like a hack but I can't think of a better
+        idea. The root problem is we can now attach ports to networks with
+        a different backend driver/ipam strategy than the network speficies.
```
roaet: Just for my education: why do ports override ipam strategies?

roaet: Nevermind I got it

quark/drivers/ironic_driver.py
morgabra: Auth should work if we turned it on, but as it sits now it's noauth.

quark/drivers/ironic_driver.py
morgabra: Didn't want to point at the local default port is all.

quark/api/extensions/ports_quark.py
morgabra: There shouldn't be. Current this field is 'instance.node' from novas point of view. Admins should be allowed to pass this in so they can manually create ports for baremetal nodes, but definitely not tenants.

roaet: Just confirming that if a non-admin user DID include this in the body it would 40X appropriately?

quark/drivers/optimized_nvp_driver.py
```diff
@@ -226,6 +226,16 @@ def _lswitch_create_optimized(self, context, network_name, nvp_id,
...
         return new_switch
 
+    def get_lswitch_ids_for_network(self, context, network_id):
+        """Public interface for fetching lswitch ids for a given network.
+
+        NOTE(morgabra) This is here because calling private methods
+        from outside the class feels wrong, and we need to be able to
+        fetch lswitch ids for use in other drivers.
+        """
+        lswitches = self._lswitches_for_network(context, network_id)
```
roaet: checking if the differences between the optimized driver and vanilla are by design

quark/drivers/ironic_driver.py
```diff
@@ -0,0 +1,542 @@
...
+
+    # client retry options
+    cfg.IntOpt("operation_retries",
+               default=3,
+               help="Number of times to attempt client operations."),
+    cfg.IntOpt("operation_delay",
+               default=0,
+               help="Base seconds to wait between client attempts."),
+    cfg.IntOpt("operation_backoff",
+               default=0,
```
mpath: You might want this to default to 1 rather than 0, see https://github.com/rackerlabs/quark/blob/c1dde7c6e3baa8ae3eeb955e6d26ece139e571ca/quark/utils.py#L172-L175

quark/drivers/ironic_driver.py
```diff
@@ -0,0 +1,542 @@
...
+    cfg.StrOpt('auth_strategy',
+               default='keystone',
+               help='Authorization strategy for connecting to neutron.'),
+
+    # client retry options
+    cfg.IntOpt("operation_retries",
+               default=3,
+               help="Number of times to attempt client operations."),
+    cfg.IntOpt("operation_delay",
+               default=0,
```
mpath: Let's bump this up to something like 3. I know we won't rely on the defaults in prod, but just in case we forget.

quark/drivers/ironic_driver.py
```diff
@@ -0,0 +1,542 @@
...
+            downstream port for any reason, the exception will be logged
+            and IronicException raised.
+        """
+        LOG.info("delete_port %s %s" % (context.tenant_id, port_id))
+        try:
+            self._delete_port(context, port_id)
+            LOG.info("deleted downstream port: %s" % (port_id))
+        except Exception:
+            LOG.error("failed deleting downstream port, it is now "
+                      "orphaned! port_id: %s" % (port_id))
```
mpath: If it's actually a 404, will it be logged as orphaned? I haven't tried running it to check.

quark/drivers/unmanaged.py
```diff
@@ -15,7 +15,9 @@
 
 from oslo_log import log as logging
 
+from quark.drivers import base
 from quark.drivers import security_groups as sg_driver
+
```
mpath: -

quark/ipam.py
```diff
@@ -995,12 +995,63 @@ def _choose_available_subnet(self, context, net_id, version=None,
         return subnets
 
 
+class IronicIpam(QuarkIpam):
+    """IPAM base class for the Ironic driver.
+
+    The idea here is that there are many small subnets created for a
+    particular segment for a provider network. And the Ironic IPAM
```
mpath: Grammar: "network. The"...

quark/plugin_views.py
```diff
@@ -207,8 +212,12 @@ def _make_port_address_dict(ip, port, fields=None):
     enabled = ip.enabled_for_port(port)
     subnet_id = ip.get("subnet_id")
     net_id = ip.get("network_id")
-    if STRATEGY.is_provider_network(net_id):
-        subnet_id = STRATEGY.get_provider_subnet_id(net_id, ip["version"])
+
+    subnet_id = ip.get("subnet_id")
```
mpath: Assigning subnet_id twice here (already assigned on line 213).

quark/api/extensions/ports_quark.py
morgabra: I'll get stripped: https://github.com/rackerlabs/quark/pull/500/files#diff-af665f1b0224068309ac23cc1f7f6e00R145

Which I think what we want.

quark/drivers/ironic_driver.py
morgabra: Indeed, although with no help from python-neutronclient. See: https://github.com/rackerlabs/quark/pull/500/files#diff-5979c993c0feddbffb6510b8a747cf05R404

quark/drivers/optimized_nvp_driver.py
morgabra: Yep, we can look up the correct lswitch for the network_id from the local db.

quark/drivers/ironic_driver.py
morgabra: Heh, I tried it, then I remembered - the pattern here was to use the conf options directly in the decorators on the functions, but I couldn't quickly figure out how to mock that out effectively so tests were taking forever.

Should I find a way to refactor that?

quark/drivers/unmanaged.py
morgabra: fixed

quark/ipam.py
morgabra: fixed

quark/plugin_views.py
morgabra: fixed

quark/drivers/ironic_driver.py
morgabra: Yeah, would be rad to be able to use a different network that got hidden behind a parent uuid (sort of like provider subnets do not).

We could dodge all this stuff if we could have some cells point to provider networks with the correct driver/ipam. It gets a little more complicated with user-created ports I guess.

quark/drivers/ironic_driver.py
mpath: Cool, thanks.

quark/tests/plugin_modules/test_ports.py
```diff
@@ -1155,11 +1283,11 @@ def test_port_diagnose_not_authorized(self):
                 self.plugin.diagnose_port(self.context, 1, [])
 
 
-class TestPortNetworkPlugin(test_quark_plugin.TestQuarkPlugin):
+class TestPortDriverSelection(test_quark_plugin.TestQuarkPlugin):
     @contextlib.contextmanager
     def _stubs(self, network=None, addr=None, mac=None,
-               compat_map=None, driver_res=None):
-        network["ipam_strategy"] = "ANY"
+               compat_map=None, driver_res=None, ipam="FOO"):
```
mpath: This may just be me, but I have an aversion to using foo/bar for anything because it's not descriptive. Is there a real word we could use here that tells a story? Just a nit.

quark/tests/plugin_modules/test_ports.py
```diff
@@ -1169,17 +1297,26 @@ def _stubs(self, network=None, addr=None, mac=None,
         # Mock out the driver registry
         foo_driver = mock.Mock()
         foo_driver.create_port.return_value = driver_res
+        foo_driver.select_ipam_strategy.return_value = "FOO"
         bar_driver = mock.Mock()
         bar_driver.create_port.return_value = driver_res
+        bar_driver.select_ipam_strategy.return_value = "BAR"
```
mpath: To continue the foo/bar nitpicking, would "FAKE_ONE" and "FAKE_TWO" work in place of foo/bar, respectively?

quark/drivers/ironic_driver.py
mpath: Do you mean like cfg.CONF.set_override(...)? Was that not working to override in the tests?

quark/drivers/ironic_driver.py
roaet: It's a bit late but you may want to have a table of these orphan ports.

quark/tests/plugin_modules/test_ports.py
roaet: Personally fine with foo/bar.

quark/tests/plugin_modules/test_ports.py
roaet: Same.

roaet:
+1

mpath:
:+1:

CrashenX:
Amazing team work here, thanks! We'll work on testing a package in preprod2-iad tomorrow.

PR #501, opened 2016-01-25
Fail mysql functional tests quickly

mpath:
Without this patch, each mysql functional test takes 100 seconds to fail.
There are ~40 tests, so 4,000 seconds waiting to fail.

With this change, all tests will fail almost immediately, in a good way.

JIRA:NCP-1784


andyhky:
Thanks for the tests on these.

mpath:
You're welcome.

roaet:
What does this do for environments when we don't want the DB retries to fail immediately?

roaet:
Nevermind i see it's only tests.

PR #503, opened 2016-01-25
Making the logging for IPAM fail ERROR

xroot88:
JIRA:NCP-1778


roaet:
Would merge but @quadewarren already did this. Oh well. TIL.

PR #504, opened 2016-01-27
Fix crash on first boot

mpath:
Rather than just crash when the "subnets" key isn't in the meta dict,
check that it is or skip the block.

JIRA:NCP-1798


roaet:
+1

roaet:
How often does compile_strategy get called? Should it evoke a warning log message if the key is missing?

mpath:
I think it only happens once at the start when it kicks off via oslo.

roaet:
+1 merge after tests

PR #506, opened 2016-01-27
bump aiclib version

morgabra:


roaet:
Eh i'll just use yours.

PR #507, opened 2016-01-28
Fix SegmentAllocationRange queries for non-elevated context

jimrollenhagen:
Add SegmentAllocationRange to the list of models that do not
filter on tenant_id to avoid blowing up in _model_query.

This is queried when a network is created, so non-admin users need to be
able to read this table, which means tenant_id is injected into the
query. However, the SegmentAllocationRange model doesn't have a tenant_id,
so we need to avoid filtering on that.

JIRA:NCP-1804

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -42,6 +42,11 @@ CONF = cfg.CONF
 ONE = "one"
 ALL = "all"
 
+_NO_TENANT_MODELS = (
+    models.PortIpAssociation,
+    models.LockHolder,
+    models.SegmentAllocationRange)
+
 
 # NOTE(jkoelker) init event listener that will ensure id is filled in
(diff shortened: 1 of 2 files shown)


quark/db/api.py
```diff
@@ -151,7 +151,8 @@ def _model_query(context, model, filters, fields=None):
             if model == models.IPAddress:
                 model_filters.append(model.used_by_tenant_id.in_(value))
             elif (model == models.PortIpAssociation or
-                  model == models.LockHolder):
+                  model == models.LockHolder or
+                  model == models.SegmentAllocationRange):
```
roaet: What do you think of making it a list:
```
untenanted_models = [models...,]
elif (model in untenanted_models):
```

roaet:
+2 after tests

PR #508, opened 2016-01-28
Pass context.roles to the IronicDriver backend.

morgabra:
This passes any roles from context.roles in the body of the port request in IronicDriver.

This is done this way, instead of also including them in the header, because the fundamental purpose of the IronicDriver is to enable external services to control the lifecycle of the actual configuration on the network separately from the lifecycle of the port object in general.

As such, we have to cache the roles if they are relevant to the configuration. It is, in my opinion, poor form to cache data associated to rest objects from stuff in the request header. (As a bonus, this makes what is happening on the request crystal clear)


jimrollenhagen:
Nice. LGTM :)

PR #511, opened 2016-02-03
Remove lockmode on mac address count

mpath:
Removes the lockmode on mac count query. This is an experimental fix
that needs to be tested before release or merge into master.

JIRA:NCP-1806


mpath:
DO NOT MERGE

roaet:
Would be reverted by #512. Closing

PR #513, opened 2016-02-11
Fix errors when throwing floating ip exceptions. Add check if port already has flip on update

alanquillin:
JIRA:NCP-1759

- Added new check to floating ip update api method to verify that the requested port does not already have a different floating IP.
    - Added test to verify the new check
- Fixed floating ip error messages which caused an unhandled exception when building the exception superclass.
- Fixed inconsistent uses of "IP" vs "Ip" in the floating ip exception names.


roaet:
LGTM.

PR #516, opened 2016-02-12
Fixed null value for the floating ip status

alanquillin:
NCP-1689

Since we do not currently set or track the status of a flip (because the flip controllers do not have a callback to update the status) I modified the view to set the status to either RESERVED (if not associated with a port) or ASSOCIATED (if associated with a port).


roaet:
LGTM

PR #518, opened 2016-02-13
Allow admin tenants to create ports on any tenant's network

brandonlogan:
JIRA: NCP-1819

This is my first attempt as getting code in quark. This will most likely be a trial and error process and I'm sure I've not done something correctly.


quark/tests/plugin_modules/test_ports.py
```diff
@@ -363,6 +363,21 @@ def test_RM9305_tenant_create_other_tenants_port(self):
             with self.assertRaises(exceptions.NotAuthorized):
                 self.plugin.create_port(self.context, port_1)
 
+    def test_RM9305_admin_tenant_create_other_tenants_port(self):
```
brandonlogan: So I wasn't sure if RM9305 was an old bug tracker issue that no longer exists in JIRA or what, but since this change was related, I kept the same naming convention.

brandonlogan:
Also, I'd like to get this in preprod1-iad ASAP so I can verify whether this works end-to-end with nova and neutron or if other things need to be changed.  What is the quickest and best way to get that done?  I could just change the code on the box itself and restart the API to test it out real quick, but that's probably not the best way to do it.

lmaycotte:
Is this a negative test? I don't think Neutron has service accounts or an admin tenant like nova has for the rackeradminapi. I think to add a port in any tenant you go through the neutron internal API that has no auth, unless I'm missing something.

brandonlogan:
this would be going through the internal API.  From what I have seen, nova will pass to neutron's internal API X-Roles header, with a list of the users role that made the request to nova.  Depending on what roles the policy is set up to recognize, it can evaluate of of the roles in the X-Roles header as an admin, and as such, the context.is_admin will evaluate to True.

This is pretty much how upstream neutron works, except they do use keystonemiddleware to inject the X-Roles when there is auth in front of the API.  Since the internal neutron API is noauth, testing would be a matter of just passing in X-Roles.

I could be wrong on this, so please let me know.

lmaycotte:
I understand the change but is it planned? QE doesn't has any knowledge of this.

brandonlogan:
No this will not require any role validation on the neutron internal API.  This just makes use of mechanisms already in place to evaluate any request with the X-Roles header.

roaet:
@brandonlogan I agree with @lmaycotte here: everyone is admin in internal neutron.

brandonlogan:
well perhaps a change in the commit message and title is warranted to be more specific and accurate.  Whenever context.is_admin is true, a port can be created on another tenant's network.  Would that suffice?  Either way, this code will still be needed so it will let port creation through.

brandonlogan:
I'm going to close this for now because I believe this will cause an unintended behavior.  Need to talk this one out more.

PR #519, opened 2016-02-22
Deferred allocation pool cache

mpath:
Defer hitting the database for the allocation pool cache until it
is actually accessed in code. This should lead to a performance
boost in some situations.

JIRA:NCP-1822


roaet:
LGTM

PR #520, opened 2016-02-24
Advancd services can create ports on any network

brandonlogan:
JIRA: NCP-1819

CLB 2 will be using this ability to connect nova instances hosting load
balancing software to tenants' isolated networks.  This is something
upstream openstack neutron allows with admin tenants.  CLB 2 will be making
these requests to the internal neutron endpoint.  Since all requests to
that endpoint are considered admin, only allowing advanced services makes
sense and is allowed upstream as well.

### DIFF
--- a/quark/plugin_modules/ports.py
+++ b/quark/plugin_modules/ports.py
@@ -39,9 +39,12 @@ STRATEGY = network_strategy.STRATEGY
 
 # HACK(amir): RM9305: do not allow a tenant to associate a network to a port
 # that does not belong to them unless it is publicnet or servicenet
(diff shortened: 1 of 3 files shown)


roaet:
Could you also make functional tests for this?

roaet:
Just stating my personal opinion that unit tests aren't particularly supercool and if I could have one type of tests it'd be functional.

brandonlogan:
Done.  I believe I've covered the possible cases with the functional tests.  I prefer them much better too.  I did not notice them at first.  Thanks.

alanquillin:
LGTM

PR #523, opened 2016-03-02
Redirect warnings

mpath:
Make output of tox tests cleaner by redirecting warnings to logging, with the added benefit of logging warnings in production, in case the edge cases tox is picking up on are being encountered for-realsies.

JIRA:NCP-1834

### DIFF
--- a/quark/tests/__init__.py
+++ b/quark/tests/__init__.py
@@ -1,0 +1,4 @@
+import logging
+
+
+logging.captureWarnings(True)


roaet:
Just gotta remark that we still should find out what these warnings are all about:

https://jira.rax.io/browse/NCP-1836

A further study into how to make these warnings errors (when unexpected) would be worth while.

roaet:
LGTM

xroot88:
+1

PR #524, opened 2016-03-02
Revert "Redirect warnings"

roaet:
Reverts rackerlabs/quark#523

After much discussion we should revert this so we can discuss better solutions.


xroot88:
+1 agreed on that NCP-1835 should be investigated

PR #525, opened 2016-03-02
Add error checking for NVP port missing edgecase

roaet:
JIRA:NCP-1520


quark/drivers/optimized_nvp_driver.py
```diff
@@ -103,6 +103,11 @@ def update_port(self, context, port_id, status=True,
 
     def delete_port(self, context, port_id, **kwargs):
         port = self._lport_select_by_id(context, port_id)
+        switch = None
```
xroot88: no point setting this as it is set on line 111

xroot88:
+1 with a minor nit

quark/drivers/optimized_nvp_driver.py
roaet: yup. forgot to remove that.

PR #527, opened 2016-03-07
Changed the ips/id/ports key for new client bug

roaet:
JIRA:NCP-1840

New client from upstream doesn't like it when the resource key differs
from what is expected.


roaet:
donut merge

PR #528, opened 2016-03-10
Added exception handlers and rollbacks to fix any resource lock contension due to failures when calling the Unicorn (Floating IP) API

alanquillin:
NCP-1848

Modified the database calls in the floating ip module to do an explicit session.rollback when an unhandled exception occurs.  There was an issue with the `with session.begin()` block that would leave session transactions open when the exception is raised and a deadlock would cause the api to lock up.  This forced the transaction to rollback and the re-raises the caught exception.


jkoelker:
LGTM

PR #529, opened 2016-03-10
Fixed floating ip quota issue

alanquillin:
NCP-1847

- Fixed issue where the quota count for floating ips was counting ALL ips associated to the tenant regardless of its address_type.

### DIFF
--- a/quark/db/api.py
+++ b/quark/db/api.py
@@ -91,7 +91,8 @@ def _model_query(context, model, filters, fields=None):
     filters = filters or {}
     model_filters = []
     eq_filters = ["address", "cidr", "deallocated", "ip_version", "service",
-                  "mac_address_range_id", "transaction_id", "lock_id"]
+                  "mac_address_range_id", "transaction_id", "lock_id",
+                  "address_type"]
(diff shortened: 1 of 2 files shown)


quark/db/api.py
```diff
@@ -1026,6 +1026,17 @@ def floating_ip_disassociate_fixed_ip(context, floating_ip):
     return floating_ip
 
 
+def floating_ip_count_all(context, filters):
+    query = context.session.query(sql_func.count(models.IPAddress.id))
+    model_filters = _model_query(context, models.IPAddress, filters)
+
+    if filters.get("address_type"):
```
jkoelker: I belieb if we add address type to the list of eq_filters here at https://github.com/alanquillin/quark/blob/FixFloatingIPQuotas/quark/db/api.py#L93
then we can just call the ip_address_count_all function if the filters and no need for a separate function.

jkoelker:
LGTM

jkoelker:
Fine travis, be like that!

jkoelker:
LGTM

PR #530, opened 2016-03-12
Removed redundant call to deallocate the floating IP address

alanquillin:
NCP-1847

Removed redundant calls to deallocate the ip address.  Once to IPAM, and the second to the data_api directly.


jkoelker:
LGTM

PR #531, opened 2016-03-14
Fixed issue where the floating ip does not get cleaned up when there is an error from the unicorn api

alanquillin:
NCP-1852

This changes the logic to allow neutron to properly cleanup the floating ip data in the neutron db, even if an error is returned from the unicorn api.  This is fine, since the unicorn api uses neutron as the source of truth, so when the ip is reused, any residual data in the unicorn system will be replaced with the new data.


jkoelker:
LGTM. Eventually we might want to retry a few times, but for now is fine.

PR #534, opened 2016-03-21
pegged TwiceRedis version to 1.0.7

tr3buchet:


tr3buchet:
this is needed because i've updated the way TwiceRedis is instantiated in version 2.0.0. Unless the version is pegged to a 1.x.x release, new virtual env builds will pull down the 2.0.0 version and quark won't be using it correctly. This is an easy small change for cherry-picking any where it is necessary and another will follow that updates quark to use TwiceRedis 2.0.0

roaet:
LGTM.

PR #537, opened 2016-03-23
Making IPPolicy.get_ip_policy_cidrs an inst method

quadewarren:
JIRA:NCP-1315


quark/db/api.py
```diff
@@ -404,7 +404,10 @@ def ip_address_reallocate_find(context, transaction_id):
 
     # TODO(amir): performance test replacing this with SQL in
     #             ip_address_reallocate's UPDATE statement
-    policy = models.IPPolicy.get_ip_policy_cidrs(subnet)
+    ip_policy_subnet = subnet["ip_policy"]
+    policy = netaddr.IPSet([])
+    if ip_policy_subnet is not None:
+        policy = ip_policy_subnet.get_cidrs_ip_set()
```
roaet: For this and the ones below it, should anything happen if the subnet is None?

quark/tests/test_db_models.py
```diff
@@ -1,40 +0,0 @@
...
-        self.assertEqual(ip_policy_rules, IPSet())
-
-    def test_get_ip_policy_cidrs_v6(self):
-        subnet = dict(id=1, ip_version=6, next_auto_assign_ip=0,
-                      cidr="fc00::/7",
-                      first_ip=334965454937798799971759379190646833152L,
-                      last_ip=337623910929368631717566993311207522303L,
-                      network=dict(ip_policy=None), ip_policy=None)
-        ip_policy_rules = models.IPPolicy.get_ip_policy_cidrs(subnet)
-        self.assertEqual(ip_policy_rules, IPSet())
```
roaet: Can't believe there was a whole file dedicated to testing the jank

quark/db/api.py
```diff
@@ -404,7 +404,10 @@ def ip_address_reallocate_find(context, transaction_id):
 
     # TODO(amir): performance test replacing this with SQL in
     #             ip_address_reallocate's UPDATE statement
-    policy = models.IPPolicy.get_ip_policy_cidrs(subnet)
+    ip_policy_subnet = subnet["ip_policy"]
+    policy = netaddr.IPSet([])
```
xroot88: I think line 408 should go into the else part of the if for readability:
```python
if ip_policy_subnet is not None:
    policy = ip_policy_subnet.get_cidrs_ip_set()
else:
    policy = netaddr.IPSet([])
```

quark/db/models.py
```diff
@@ -327,7 +327,10 @@ def allocation_pools(self):
             pools = json.loads(_cache)
             return pools
         else:
-            ip_policy_cidrs = IPPolicy.get_ip_policy_cidrs(self)
+            ip_policy = self["ip_policy"]
```
xroot88: same as in prev comment

quark/ipam.py
```diff
@@ -452,7 +452,10 @@ def _allocate_from_subnet(self, context, net_id, subnet,
                                                  port_id=port_id,
                                                  ip_address=ip_address)))
 
-        ip_policy_cidrs = models.IPPolicy.get_ip_policy_cidrs(subnet)
+        ip_policy_subnet = subnet["ip_policy"]
```
xroot88: same as above

quark/db/api.py
```diff
@@ -404,7 +404,15 @@ def ip_address_reallocate_find(context, transaction_id):
 
     # TODO(amir): performance test replacing this with SQL in
     #             ip_address_reallocate's UPDATE statement
-    policy = models.IPPolicy.get_ip_policy_cidrs(subnet)
+    ip_policy_subnet = None
```
xroot88: It looks like this fragment is repeated 4 times - just wondering why not make it a function and put it in some common file?

quadewarren: That is actually what we were replacing. Originally, this was a function attached to the IPPolicy class, but we made it an instance method, so these checks became needed. We could wrap it in a function, but I don't think this snippet of code belongs in models.py and if we created a new file to contain it, we might as well shove other common functions, too. I don't think that would make the code work more clearly, however.

roaet:
LGTM +1

🐣

quark/ipam.py
```diff
@@ -518,7 +526,15 @@ def _allocate_from_v6_subnet(self, context, net_id, subnet,
...
-            ip_policy_cidrs = models.IPPolicy.get_ip_policy_cidrs(subnet)
+            ip_policy_subnet = None
+            if subnet is not None:
+                ip_policy_subnet = subnet["ip_policy"]
+
+            if ip_policy_subnet is not None:
+                ip_policy_cidrs = ip_policy_subnet.get_cidrs_ip_set()
+            else:
+                ip_policy_cidrs = netaddr.IPSet([])
+
```
xroot88: Maybe shorten this block a little to something like this?:
```python
            if subnet and subnet['ip_policy']:
                ip_policy_cidrs = subnet['ip_policy'].get_cidrs_ip_set()
            else:
                ip_policy_cidrs = netaddr.IPSet([])
```

xroot88:
+1 - check my comment to make code more readable, otherwise, looks good

quark/ipam.py
quadewarren: That works.

PR #538, opened 2016-03-29
Add Scaling IP API extension

brandonlogan:
JIRA: NCP-1774

The current Floating IPs extension allows a one-to-one association of an IP
address to a port's fixed ip address.  This is aligned with neutron's
Floating IP API and should not be changed.  Scaling IPs is essentially a
floating IP with a one-to-many association of an IP address to a ports'
fixed ip addresses.  Due to this simple, but fundamentally different, change
a new extension and API resource makes the most sense.

A new database enum value is needed to differentiate between a floating
and scaling IP address.  This ensures that when a user retrieves a list of
floating IPs, scaling IPs are not returned in that list.  It also ensures the
inverse as well.

The consolidation of code to reduce repeated code has been strived for, but
it could always be better.  The current state tries to strike a balance between
complete DRY and drastically changing the exisitng floating IP code.


alanquillin:
LGTM

PR #539, opened 2016-03-29
Adjusting Quark to NOOP on isonet vifs in FLIP enabled environments

Anonymike:


roaet:
Are there any situations where something in NVP is not an isolated network?

Could you please squash your PR to a single commit (unless you're not done yet).

roaet:
Ah nevermind. I see that you're on the agent. So the vif doesn't know where it came from.

roaet:
Can this be moiged?

PR #540, opened 2016-04-18
Allowing subnet allocation pool shrinking

roaet:
JIRA:NCP-1869

To allow shrinking set allow_allocation_pool_update = True.

- added new configuration allow_allocation_pool_growth (defaults to
  False)
- few old tests needed new overrides for new cfg
- without configuration changes this should not deviate from current
  behavior
- allocation pool growth is actually undefined, but this allows the DB
  portion to change

### DIFF
--- a/quark/plugin_modules/subnets.py
+++ b/quark/plugin_modules/subnets.py
@@ -41,8 +41,12 @@ STRATEGY = network_strategy.STRATEGY
 quark_subnet_opts = [
     cfg.BoolOpt('allow_allocation_pool_update',
                 default=False,
-                help=_('Controls whether or not to allow allocation_pool'
-                       'updates'))
+                help=_('Controls whether or not to allow allocation_pool '
+                       'updates')),
+    cfg.BoolOpt('allow_allocation_pool_growth',
+                default=False,
+                help=_('Controls whether or not to allow allocation_pool '
+                       'growing. Otherwise shrinking is only allowed'))
 ]
 
 CONF.register_opts(quark_subnet_opts, "QUARK")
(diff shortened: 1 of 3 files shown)


quark/plugin_modules/subnets.py
```diff
@@ -226,6 +230,21 @@ def create_subnet(context, subnet):
...
+    for rng in original_pool._alloc_pools:
+        ori_set.add(netaddr.IPRange(rng['start'], rng['end']))
+
+    # create IPSet for net pool
+    new_set = netaddr.IPSet()
+    for rng in new_pool._alloc_pools:
+        new_set.add(netaddr.IPRange(rng['start'], rng['end']))
+
+    # we are growing if ori UNION new != ori
+    return ori_set | new_set != ori_set
```
andyhky: can't we just return len(ori_set) > len(new_set)? I think that logic is easier to grok

roaet: That isn't necessarily going to cover the expansion of the boundaries of the pool.

xroot88: If I understood this correctly, will the pool be only growing when the new_set is a superset of the ori_set?
Like this:
```python
return new_set.issuperset(ori_set)
```

quark/plugin_modules/subnets.py
```diff
@@ -267,6 +286,15 @@ def update_subnet(context, id, subnet):
...
                                                           allocation_pools)
+            original_pools = subnet_db.allocation_pools
+            ori_pools = allocation_pool.AllocationPools(subnet_db["cidr"],
+                                                        original_pools)
+            # Check if the pools are growing or shrinking
+            is_growing = _pool_is_growing(ori_pools, alloc_pools)
+            if not CONF.QUARK.allow_allocation_pool_growth and is_growing:
+                raise n_exc.BadRequest(
+                    resource="subnets",
+                    msg="Allocation pools may not be updated to be larger.")
```
xroot88: A suggestion to include with the message that this is due to the config setting.
```python
msg="Allocation pools may not be updated to be larger due to config settings"
```

quark/plugin_modules/subnets.py
roaet: No because one can 'expand the boundaries' of the original set and not contain all of the things in the original set. issuperset would fail at that point.

quark/plugin_modules/subnets.py
roaet: +1

quark/plugin_modules/subnets.py
roaet: To further describe:

```
012345
--|||- <- old
-|||-- <- new
```

roaet: A union provides the 'bitmasking' effect that I am looking for. If the mask of the original fully encompasses the other one, then it's fine.

Alternatively I guess we could say:

```
return not ori_set.issuperset(new_set)
```

roaet:
If A | B = A, then A has to be a super set of A. Right? Right?

xroot88:
+2

PR #541, opened 2016-04-20
Adding the device id to port delete logging

roaet:
JIRA:NCP-1398

It only does this if the device_id is available (or none). No tests as
it is just logging.


xroot88:
+2

PR #542, opened 2016-04-20
Just skips subnets that are None during ipam

roaet:
JIRA:NCP-1304

This is an incredibly rare and non-impacting situation. All it does is
add a single iteration to the loop and creates weird log messages. Since
it doesn't really do anything to add this continue I'm not going to
bother to create a test (since the desired effect is to do what it does
right now).

Again, this primarily will just try to make logging more clear in the
future.


xroot88:
+2

PR #543, opened 2016-04-21
Fix disassociating ports from SCIPs on port delete

brandonlogan:
JIRA: NCP-1909

Whenever a port is deleted that was associated with a scaling IP, the scaling
IP needs to only disassociate itself from that port and it's fixed ips.  This
fixes a bug where the scaling IP was disassociating itself from all ports
whenever one of its ports were deleted.


alanquillin:
LTGTM

PR #549, opened 2016-05-05
Fixed IPAddressInUse on allocating deallocated ip

quadewarren:
JIRA:NCP-1898

### DIFF
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -565,8 +565,12 @@ class QuarkIpam(object):
                      new_addresses=new_addresses, ip_address=ip_address)))
 
         subnets = subnets or []
+        allocated_ips = [ip.get("address_readable") for ip in new_addresses]
         for subnet in subnets:
             if not subnet:
+                continue
+
+            if str(ip_address) in allocated_ips:
                 continue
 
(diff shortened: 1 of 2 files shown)


quark/ipam.py
```diff
@@ -565,10 +565,14 @@ def _allocate_ips_from_subnets(self, context, new_addresses, net_id,
                      new_addresses=new_addresses, ip_address=ip_address)))
 
         subnets = subnets or []
+        allocated_ips = [ip.get("address_readable") for ip in new_addresses]
         for subnet in subnets:
             if not subnet:
                 continue
 
+            if str(ip_address) in allocated_ips:
+                continue
```
xroot88: Here, if the ip_address was found in the new_addresses list, we return new_addresses unmodified.
Is this intended?

quadewarren: This behavior is unchanged from the previous code. new_addresses actually doesn't appear to serve a purpose in this function. I've been seeing a lot of code that appears unused, so it looks like this code was gutted.

roaet:
LGTM +1

🐣

xroot88:
+1

PR #550, opened 2016-05-11
Added qualified exception to ipam

quadewarren:
JIRA:NCP-1912

### DIFF
--- a/quark/tests/functional/test_ipam.py
+++ b/quark/tests/functional/test_ipam.py
@@ -22,6 +22,7 @@ from neutron_lib import exceptions as n_exc
 
 from quark.db import api as db_api
 from quark.db import models
+from quark import exceptions as q_exc
 import quark.ipam
 
 # import below necessary if file run by itself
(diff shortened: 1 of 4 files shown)


quark/tests/functional/test_ipam.py
```diff
@@ -182,3 +183,43 @@ def test_subnets_get_marked_as_full_retroactively(self):
...
+            sub_mod = db_api.subnet_create(self.context, **subnet)
+            # NOTE(asadoughi): update after cidr constructor has been invoked
+            db_api.subnet_update(self.context,
+                                 sub_mod,
+                                 next_auto_assign_ip=next_ip)
+        yield net_mod
+        with self.context.session.begin():
+            db_api.subnet_delete(self.context, sub_mod)
+            db_api.network_delete(self.context, net_mod)
+
```
roaet: I see you are attempting to reallocate a single provided ipv4 and letting the ipv6 case run through (i think?). Can you make very similar tests that iterate through the other provided possibilities (just ipv6, both ipv6 and ipv4)?

roaet:
LGTM other than test question

quark/tests/functional/test_ipam.py
quadewarren: Yeah, that makes sense. You are only talking about different test cases and not different IPAM strategies, right?

roaet:
LGTM +1 ⛄

xroot88:
+1

PR #552, opened 2016-05-11
Fixing the travis tests by keeping the facade

roaet:


quark/tests/functional/base.py
```diff
@@ -16,13 +16,11 @@ def setUp(self):
         configure_mappers()
         # Must set the neutron's facade to none before each test
```
quadewarren: Probably need to lose these comments.

quadewarren:
LGTM

PR #554, opened 2016-05-13
undefered pools during subnet select

roaet:
JIRA:NCP-1907

This will cause a 500 ObjectDeleted exception to be raised should the
subnet be deleted during the get on the defered column


xroot88:
+2

quadewarren:
LGTM, even though you had to explain it, haha

PR #555, opened 2016-05-19
Fix bug with list of scaling IPs

brandonlogan:
JIRA:NCP-1938

If a scaling IP was not associated with any ports then it
would fail to be converted to a dictionary, causing 500 on
a GET request to /scalingips


alanquillin:
LGTM

PR #556, opened 2016-06-01
Optimized reallocation for BOTH strategy

quadewarren:
JIRA:NCP-1930

Reallocation algorithm would originally attempt to reallocate an ip as
both an ipv6 and an ipv4. This change sets the algorithm to only attempt
to reallocate a given ip based on the given ip version or to attempt
to reallocate an ipv4 given the parameters.

### DIFF
--- a/quark/ipam.py
+++ b/quark/ipam.py
@@ -980,13 +980,13 @@ class QuarkIpamBOTH(QuarkIpam):
                                  reuse_after, version=None,
                                  ip_address=None, segment_id=None,
                                  subnets=None, **kwargs):
-        both_versions = []
-        for ver in (4, 6):
-            address = super(QuarkIpamBOTH, self).attempt_to_reallocate_ip(
-                context, net_id, port_id, reuse_after, ver, ip_address,
-                segment_id, subnets=subnets, **kwargs)
-            both_versions.extend(address)
-        return both_versions
+        ip_address_version = 4 if not ip_address else ip_address.version
+        # NOTE(quade): We do not attempt to reallocate ipv6, so just return
+        if ip_address_version == 6:
+            return []
+        return super(QuarkIpamBOTH, self).attempt_to_reallocate_ip(
+            context, net_id, port_id, reuse_after, ip_address_version,
+            ip_address, segment_id, subnets=subnets, **kwargs)
 
     def _choose_available_subnet(self, context, net_id, version=None,
                                  segment_id=None, ip_address=None,


quark/ipam.py
```diff
@@ -981,11 +981,11 @@ def attempt_to_reallocate_ip(self, context, net_id, port_id,
...
-        for ver in (4, 6):
-            address = super(QuarkIpamBOTH, self).attempt_to_reallocate_ip(
-                context, net_id, port_id, reuse_after, ver, ip_address,
-                segment_id, subnets=subnets, **kwargs)
-            both_versions.extend(address)
+        ip_address_version = 4 if not ip_address else ip_address.version
+        address = super(QuarkIpamBOTH, self).attempt_to_reallocate_ip(
+            context, net_id, port_id, reuse_after, ip_address_version,
+            ip_address, segment_id, subnets=subnets, **kwargs)
+        both_versions.extend(address)
```
mpath: both_versions will only ever contain one item now, correct?

quadewarren: Yes, I kept it this way since the caller expected a list back. The strange thing is that it would only receive a list with one item in it back, since we don't attempt to reallocate an ipv6.

mpath: then this: both_versions.extend(address)
could be just:
both_versions = [address], the dot-call is just going to slow it down and isn't needed

quadewarren: Well, address can also be returned as [] if the given ip IS an ipv6, which would result in both_versions = [[]] instead of just [] with that code change, but it is avoided with extend.

mpath: would [[]] break anything?

mpath:
👍

quark/ipam.py
```diff
@@ -980,13 +980,10 @@ def attempt_to_reallocate_ip(self, context, net_id, port_id,
...
                                  ip_address=None, segment_id=None,
                                  subnets=None, **kwargs):
-        both_versions = []
-        for ver in (4, 6):
-            address = super(QuarkIpamBOTH, self).attempt_to_reallocate_ip(
-                context, net_id, port_id, reuse_after, ver, ip_address,
-                segment_id, subnets=subnets, **kwargs)
-            both_versions.extend(address)
-        return both_versions
+        ip_address_version = 4 if not ip_address else ip_address.version
```
xroot88: if this is only for ipv4, do we need this?

quadewarren: Technically, no. ipv6 will never attempt to reallocate, so I could change it even further so that if it's an ipv6, it would just return an empty list and skip some calls all together.

stajkowski:
LGTM

PR #558, opened 2016-06-02
Ip notifications

xroot88:
JIRA:NCP-1871

First commit for additional IP billing notifications.

### DIFF
--- a/quark/billing.py
+++ b/quark/billing.py
@@ -1,0 +1,297 @@
+# Copyright 2016 Openstack Foundation
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+"""Calculations for different cases for additional IP billing
+A - allocated
+D - deallocated
+| - start or end of the billing period, typically 24 hrs
+
+A |     | D
+  | A   | D
+A | D   |
+  | A D |
+
+---IP CURRENTLY ALLOCATED=has tenant_id (send usage at end of period)---
+case 1 (allocated before period began, deallocated after period ended)
+  send usage as 24 hours
+case 2 (allocated during period, deallocated after period ended)
+  send usage as timedelta(period_end - allocated_at)
+
+---IP DEALLOCATED DURING PERIOD (send usage at deallocation time)---
+case 3 (allocated before period began, deallocated during period)
+  send usage as timedelta(deallocated_at - period_start)
+case 4 (allocated during period, deallocated during period)
+  send usage as timedelta(deallocated_at - allocated_at)
+
+NOTE: notifications are issued at these times:
+    case1 and case2 - this script runs nightly and processes the entries
+    case3 and case4 - the notifications are sent immediately when the ip
+                      is deallocated
+NOTE: assumes that the beginning of a billing cycle is midnight.
+"""
+
+import datetime
+from neutron.common import rpc as n_rpc
+from oslo_log import log as logging
+from sqlalchemy import and_, or_, null
+from quark.db import models
+
+from quark import network_strategy
+
+LOG = logging.getLogger(__name__)
+
+PUBLIC_NETWORK_ID = network_strategy.STRATEGY.get_public_net_id()
+
+# NOTE: this will most likely go away to be done in yagi
+EVENT_TYPE_2_CLOUDFEEDS = {
+    'ip.exists': 'USAGE',
+    'ip.add': 'CREATE',
+    'ip.delete': 'DELETE',
+    'ip.associate': 'UP',
+    'ip.disassociate': 'DOWN'
+}
+
+
+def do_notify(context, event_type, payload):
+    """Generic Notifier.
+
+    Parameters:
+        - `context`: session context
+        - `event_type`: the event type to report, i.e. ip.usage
+        - `payload`: dict containing the payload to send
+    """
+    LOG.debug('IP_BILL: notifying {}'.format(payload))
+
+    notifier = n_rpc.get_notifier('network')
+    notifier.info(context, event_type, payload)
+
+
+def notify(context, event_type, ipaddress, send_usage=False):
+    """Method to send notifications.
+
+    We must send USAGE when a public IPv4 address is deallocated or a FLIP is
+    associated.
+    Parameters:
+        - `context`: the context for notifier
+        - `event_type`: the event type for IP allocate, deallocate, associate,
+        disassociate
+        - `ipaddress`: the ipaddress object to notify about
+    Returns:
+        nothing
+    Notes: this may live in the billing module
+    """
+    # ip.add needs the allocated_at time.
+    # All other events need the current time.
+    ts = ipaddress.allocated_at if event_type == 'ip.add' else _now()
+    payload = build_payload(ipaddress, event_type, event_time=ts)
+
+    # Send the notification with the payload
+    do_notify(context, event_type, payload)
+
+    # When we deallocate an IP or associate a FLIP we must send
+    # a usage message to billing.
+    # In other words when we supply end_time we must send USAGE to billing
+    # immediately.
+    # Our billing period is 24 hrs. If the address was allocated after midnight
+    # send the start_time as as. If the address was allocated yesterday, then
+    # send midnight as the start_time.
+    # Note: if allocated_at is empty we assume today's midnight.
+    if send_usage:
+        if ipaddress.allocated_at is not None and \
+           ipaddress.allocated_at >= _midnight_today():
+            start_time = ipaddress.allocated_at
+        else:
+            start_time = _midnight_today()
+        payload = build_payload(ipaddress,
+                                'ip.exists',
+                                start_time=start_time,
+                                end_time=ts)
+        do_notify(context, 'ip.exists', payload)
+
+
+def build_payload(ipaddress,
+                  event_type,
+                  event_time=None,
+                  start_time=None,
+                  end_time=None):
+    """Method builds a payload out of the passed arguments.
+
+    Parameters:
+        `ipaddress`: the models.IPAddress object
+        `event_type`: USAGE,CREATE,DELETE,SUSPEND,or UNSUSPEND
+        `start_time`: startTime for cloudfeeds
+        `end_time`: endTime for cloudfeeds
+    Returns a dictionary suitable to notify billing.
+    Message types mapping to cloud feeds for references:
+        ip.exists       - USAGE
+        ip.add          - CREATE
+        ip.delete       - DELETE
+        ip.associate    - UP
+        ip.disassociate  - DOWN
+    Refer to: http://rax.io/cf-api for more details.
+    """
+    # This is the common part of all message types
+    payload = {
+        'event_type': unicode(EVENT_TYPE_2_CLOUDFEEDS[event_type]),
+        'tenant_id': unicode(ipaddress.used_by_tenant_id),
+        'ip_address': unicode(ipaddress.address_readable),
+        'subnet_id': unicode(ipaddress.subnet_id),
+        'network_id': unicode(ipaddress.network_id),
+        'public': True if ipaddress.network_id == PUBLIC_NETWORK_ID else False,
+        'ip_version': int(ipaddress.version),
+        'ip_type': unicode(ipaddress.address_type),
+        'id': unicode(ipaddress.id)
+    }
+
+    # Depending on the message type add the appropriate fields
+    if event_type == 'ip.exists':
+        if start_time is None or end_time is None:
+            raise ValueError('IP_BILL: {} start_time/end_time cannot be empty'
+                             .format(event_type))
+        payload.update({
+            'startTime': unicode(convert_timestamp(start_time)),
+            'endTime': unicode(convert_timestamp(end_time))
+        })
+    elif event_type == 'ip.add':
+        if event_time is None:
+            raise ValueError('IP_BILL: {}: event_time cannot be NULL'
+                             .format(event_type))
+        payload.update({
+            'eventTime': unicode(convert_timestamp(event_time)),
+        })
+    elif event_type == 'ip.delete':
+        if event_time is None:
+            raise ValueError('IP_BILL: {}: event_time cannot be NULL'
+                             .format(event_type))
+        payload.update({
+            'eventTime': unicode(convert_timestamp(event_time))
+        })
+    elif event_type == 'ip.associate' or event_type == 'ip.disassociate':
+        if event_time is None:
+            raise ValueError('IP_BILL: {}: event_time cannot be NULL'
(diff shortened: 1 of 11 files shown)


quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+case 3 (allocated before period began, deallocated during period)
+  send usage as timedelta(deallocated_at - period_start)
+case 4 (allocated during period, deallocated during period)
+  send usage as timedelta(deallocated_at - allocated_at)
+
+NOTE: notifications are issued at these times:
+    case1 and case2 - this script runs nightly and processes the entries
+    case3 and case4 - the notifications are sent immediately when the ip
+                      is deallocated
+NOTE: assumes that the beginning of a billing cycle is midnight.
```
quadewarren: Midnight in what time zone? For that region or is it static against UTC?

quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+    ts_now = now()
+
+    # Build payloads based on the message type and supply the correct start/end
+    # or event times.
+    if event_type == 'ip.add':
+        # allocated_at is more precise than now(), so we use allocated_at
+        payload = build_payload(ipaddress,
+                                event_type,
+                                event_time=ipaddress.allocated_at)
+    elif event_type == 'ip.delete':
```
quadewarren: Since  they're doing the same thing, could we do "elif event_type in EVENT_TYPE_2_CLOUDFEEDS.keys()" and then follow up with an else?

quark/billing.py
roaet: If it timezone related we'd need to at configuration and deployment info for the nodes.

quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+
+from oslo_log import log as logging
+
+from sqlalchemy import and_, or_, null
+
+from quark.db import models
+
+LOG = logging.getLogger(__name__)
+
+PUBLIC_NETWORK_ID = u'00000000-0000-0000-0000-000000000000'
```
roaet: nit: technically this UUID may not be the real public network UUID. It'd be best to get the UUID from the network_strategy

quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+    """Generic Notifier.
+
+    Parameters:
+        - `context`: session context
+        - `event_type`: the event type to report, i.e. ip.usage
+        - `payload`: dict containing the payload to send
+    """
+    LOG.debug('IP_BILL: notifying {}'.format(payload))
+
+    notifier = n_rpc.get_notifier('network')
```
roaet: "network" seems pretty generic. Is this a pre-existing condition or did we come up with it?

quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+        payload.update({
+            'startTime': unicode(convert_timestamp(start_time)),
+            'endTime': unicode(convert_timestamp(end_time))
+        })
+    elif event_type == 'ip.add':
+        if event_time is None:
+            raise ValueError('IP_BILL: {}: event_time cannot be NULL'
+                             .format(event_type))
+        payload.update({
+            'eventTime': unicode(convert_timestamp(event_time)),
```
quadewarren: Is the trailing comma supposed to be there? I think python will interpret that as a creation of a tuple.

quark/billing.py
roaet: I sort of agree with @quadewarren here, but different suggestion:

The only change is the event_time portion. Why not just branch to set that and have the same build_payload call?

If you want to ignore the ip.exists key do that explicitly and first since that is the *only* think you are ignoring.

quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+        `start_time`: startTime for cloudfeeds
+        `end_time`: endTime for cloudfeeds
+    Returns a dictionary suitable to notify billing.
+    Message types mapping to cloud feeds for references:
+        ip.exists       - USAGE
+        ip.add          - CREATE
+        ip.delete       - DELETE
+        ip.associate    - UP
+        ip.disassociate  - DOWN
+    http://docs-internal-staging.rackspace.com/cloud-feeds/api/v1.0/feeds-publishers-guide/feeds-publishers-guide-latest.pdf
```
roaet: nit: can this url be tiny-ified

quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+from sqlalchemy import and_, or_, null
+
+from quark.db import models
+
+LOG = logging.getLogger(__name__)
+
+PUBLIC_NETWORK_ID = u'00000000-0000-0000-0000-000000000000'
+
+EVENT_TYPE_2_CLOUDFEEDS = {
+    'ip.exists': 'USAGE',
```
roaet: stink: these strings are all over the place. Can you make constants and use those? (ip.exists, etc)

quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+    Seconds that occurred since midnight till the timestamp ts.
+    Parameters:
+        - `ts`: timestamp in datetime format
+        Returns:
+            seconds elapsed since midnight till ts
+    """
+    return (ts - ts.replace(hour=0, minute=0, second=0)).seconds
+
+
+def make_case2(context):
```
quadewarren: Is this still needed?

quark/tools/billing.py
```diff
@@ -0,0 +1,107 @@
...
+    hour = int(sys.argv[1])
+    minute = int(sys.argv[2])
+    # Read the config file and get the admin context
+    config_opts = ['--config-file', '/etc/neutron/neutron.conf']
+    config.init(config_opts)
+    config.setup_logging()
+    context = neutron_context.get_admin_context()
+
+    # Here one can make case 2 entry in the db
+    # billing.make_case2(context)
```
quadewarren: Is this still needed?

quark/billing.py
roaet: It is in curly braces and would be considered a dict. Trailing commas in dicts are fine.

quark/ipam.py
```diff
@@ -480,6 +480,10 @@ def _allocate_from_subnet(self, context, net_id, subnet,
                     port_id=port_id,
                     address_type=kwargs.get('address_type', ip_types.FIXED))
                 address["deallocated"] = 0
+                # alexm: instead of notifying from here we notify from
+                # allocate_ip_address() when it's clear that the IP
+                # allocation was successful
+                # notify(context, 'ip.add', address)
```
quadewarren: Still needed or preserve for posterity?

quark/billing.py
```diff
@@ -0,0 +1,369 @@
...
+    # this will force period_end to store something like:
+    # datetime.datetime(2016, 5, 19, 23, 59, 59, 999999)
+    # instead of:
+    # datetime.datetime(2016, 5, 20,  0,  0,  0,      0)
+    period_end -= datetime.timedelta(seconds=1)
+
+    return (period_start, period_end)
+
+
+def midnight_from_datetime(ts):
```
roaet: Based on this function (and the one below) it looks like billing is not timezone sensitive. I believe it makes sense for it to be?

quark/billing.py
roaet: If it is needed I believe it should be either prefixed with _ or put into the testing file.

quark/tests/test_billing.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class QuarkBillingBaseTest(test_base.TestBase):
+    def setUp(self):
+        super(QuarkBillingBaseTest, self).setUp()
+
+TENANT_ID = u'12345'
+IP_ID = 'ff8826b0-671d-46df-bce1-784f1bc2d1b6'
+IP_READABLE = '1.1.1.1'
+SUBNET_ID = '1a488e0b-4ab3-4299-b6af-64a144fd7198'
```
roaet: nit: this UUID is not obviously a test UUID. Can you make it something more fake looking?

quark/tests/test_billing.py
```diff
@@ -0,0 +1,145 @@
...
+from quark.db.models import IPAddress
+from quark.tests import test_base
+
+
+class QuarkBillingBaseTest(test_base.TestBase):
+    def setUp(self):
+        super(QuarkBillingBaseTest, self).setUp()
+
+TENANT_ID = u'12345'
+IP_ID = 'ff8826b0-671d-46df-bce1-784f1bc2d1b6'
```
roaet: nit: this UUID is not obviously a test UUID. Can you make it something more fake looking?

quadewarren:
Quite the monster. LGTM other than comments.

quark/tools/billing.py
```diff
@@ -0,0 +1,107 @@
...
+from neutron.common import config
+from neutron import context as neutron_context
+
+from pprint import pprint as pp
+from quark import billing
+from quark.db import models
+
+
+def main():
+    if len(sys.argv) < 3:
```
roaet: nit: in general I dislike weak command line tools. I recommend checking out click to make this a bit more safe.

quark/billing.py
xroot88: everything is UTC as everywhere in quark

quark/billing.py
xroot88: Right, this code was changing almost on a daily basis a couple weeks ago. I will compress it.

quark/billing.py
xroot88: everything is in utc

quark/billing.py
xroot88: thank you, some of these are remnants of older versions. I will clean up.

quark/ipam.py
xroot88: edited comment, removed the commented out call

quark/tests/test_billing.py
xroot88: faked

quark/tests/test_billing.py
xroot88: faked now

quark/tools/billing.py
xroot88: re-implemented using click.

quark/tools/billing.py
xroot88: removed

quark/billing.py
roaet: So that means the billing cycle stuff will occur at 6pm in dfw?

quark/billing.py
roaet: If we're trying to be consistent then using network strategy is important.

quark/billing.py
```diff
@@ -157,7 +140,7 @@ def build_payload(ipaddress,
         ip.delete       - DELETE
         ip.associate    - UP
         ip.disassociate  - DOWN
-    http://docs-internal-staging.rackspace.com/cloud-feeds/api/v1.0/feeds-publishers-guide/feeds-publishers-guide-latest.pdf
+    Refer to: http://rax.io/cf-api for more details.
```
roaet: Aww yea +1

quark/billing.py
```diff
@@ -0,0 +1,321 @@
...
+    return ts.replace(hour=0, minute=0, second=0)
+
+
+def midnight_today():
+    return datetime.datetime.utcnow().replace(hour=0,
+                                              minute=0,
+                                              second=0)
+
+
+def make_case2(context):
```
quadewarren: Can this method be shoved into a test file as part of a setUp?

quark/tools/billing.py
```diff
@@ -27,30 +27,22 @@
...
-        sys.stderr.write(' the script prints')
-        sys.stderr.write(' the messages to stdout and exits\n')
-        sys.stderr.write('Ex: "{0} 0 0 --notify=true"'.format(sys.argv[0]))
-        sys.stderr.write(' for a period starting at midnight')
-        sys.stderr.write(' and send billing msgs\n')
-        return 1
-    hour = int(sys.argv[1])
-    minute = int(sys.argv[2])
+@click.command()
+@click.option('--notify', default='False',
```
roaet: seems like you want a boolean flag:
http://click.pocoo.org/5/options/#boolean-flags

quark/billing.py
xroot88: missed this one. Sounds good.

quark/billing.py
roaet: Aye. I think it should be in the test file if it is there for testing porpoises (or at least prefixed with _)

quark/billing.py
xroot88: my understanding is that we report everything in UTC to cloud feeds. The consumer of this data will decide how to bill exactly.

quark/billing.py
xroot88: 'network' was pre-existing. I am re-using. This name may change when I am done with the Yagi part.

quark/billing.py
xroot88: fixed using network_strategy

quark/billing.py
xroot88: I'd like to keep it this way until the yagi part is done. This translation may [most likely] go away as yagi will do it's own translation. I put a NOTE: on it

quark/billing.py
xroot88: re-arranged this and added an _ for local methods

quark/billing.py
roaet: So will the cronjob on the 'node that performs the daily summary' happen at midnight for the region?

requirements.txt
```diff
@@ -1,5 +1,6 @@
 SQLAlchemy<1.1.0,>=1.0.10
 alembic==0.8.2
+click>=6.6
```
mpath: Should we depend on a 3rd party for what this is doing? Looks like just sys.argv and print.

quark/billing.py
xroot88: Yes, the cronjob has to run some time after midnight.

quark/tools/billing.py
```diff
@@ -0,0 +1,121 @@
...
+    ind = random.randint(0, len(ip_list) - 1)
+    address = ip_list[ind]
+    address.allocated_at = datetime.datetime.utcnow() -\
+        datetime.timedelta(days=1)
+    context.session.add(address)
+    context.session.flush()
+
+
+@click.command()
+@click.option('--notify', default='False',
```
roaet: Still think --notify should be:

http://click.pocoo.org/5/options/#boolean-flags

quark/billing.py
roaet: ok. That makes sense. Still would need some deployment stuff for midnight per regions then.

quark/network_strategy.py
```diff
@@ -98,5 +98,12 @@ def get_network_for_subnet(self, subnet_id):
             return None
         return self.subnet_strategy.get(subnet_id)["network_id"]
 
+    def get_public_net_id(self):
+        """Returns the public net id"""
+        for id, net_params in self.strategy.iteritems():
+            if net_params['bridge'] == 'publicnet':
```
roaet: can this 'publicnet' come from a config and default to 'publicnet'

quark/tests/test_billing.py
```diff
@@ -0,0 +1,161 @@
...
+from quark import billing
+from quark.db.models import IPAddress
+from quark import network_strategy
+from quark.tests import test_base
+
+
+class QuarkBillingBaseTest(test_base.TestBase):
+    def setUp(self):
+        super(QuarkBillingBaseTest, self).setUp()
+        self.strategy = {"public_network": {"bridge": "publicnet",
```
roaet: If you make 'publicnet' a config thing, which I recommended, this would need to be updated.

quadewarren:
LGTM +1

roaet:
+1. @mpath you ok with it?

mpath:
Just needs to be squarshed again.

PR #564, opened 2016-06-19
don't allow port with gateway ip on flip associate

evanscottgray:
before we associate a floating ip, we need to verify that a port on the network does not exist that is assigned a fixed ip that matches the gateway ip of the floating ip subnet


quadewarren:
This repo has been deprecated. We're now here: https://github.com/openstack/quark

You're going to have to use Gerrit...

roaet:
Aye. This is no longer a valid repo.

roaet:
Well I guess you can use it for travis checks?

